PR #21, opened 2026-05-08
Add OpenAI-compatible embedder provider

ihubanov:
## Motivation

Enables SAGE deployments to share an existing OpenAI-compatible embeddings stack (vLLM, OpenAI, LiteLLM, etc.) instead of running Ollama or relying on the hash-based default. Particularly valuable for multilingual content where stronger embedders (e.g. gte-Qwen2-1.5B-instruct, BAAI bge-m3) outperform the hash default by a wide margin.

## What changes

- New `embedding.provider` value: `openai-compatible`
- Adds `OpenAICompatibleClient` alongside the existing hash and ollama providers
- Calls `POST {base_url}/v1/embeddings` with the standard OpenAI embeddings request shape
- Configurable dimension (validated against response), optional bearer auth via `embedding.api_key`
- Optional `Named` interface so the REST `/v1/embed` info handler reports the correct provider name
- Unit tests for the new provider (httptest.NewServer mock; happy path, auth header, error cases)

## Compatibility

- Existing `hash` and `ollama` providers unaffected — additive only.
- BFT app-validator semantics unchanged. Embedder swap is below the validator layer.
- License: Apache 2.0 (matches SAGE).

## Tested

- `go test ./...` clean
- Smoke-tested in a SAGE Personal-mode container against vLLM running gte-Qwen2-1.5B-instruct (1536-dim, multilingual). Prose-only recall queries that returned 0 hits on the hash embedder return the expected fact on the new provider.

## Configuration example

```yaml
embedding:
  provider: openai-compatible
  base_url: https://api.openai.com  # or vLLM, LiteLLM, your own gateway
  model: text-embedding-3-small
  api_key: sk-...
  dimension: 1536
```

### DIFF
--- a/api/rest/embed_handler.go
+++ b/api/rest/embed_handler.go
@@ -2,6 +2,8 @@ package rest
 
 import (
 	"net/http"
+
+	"github.com/l33tdawg/sage/internal/embedding"
 )
 
 // EmbedRequest is the request body for POST /v1/embed.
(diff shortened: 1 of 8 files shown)


ihubanov:
Pulled this back — caught a few things I want to clean up before you spend time on it. Commit format wasn't matching SAGE's conventional-commits style, body had some marketing fluff, wanted to add an explicit ABCI-determinism reassurance line. Reopening shortly 🙏

PR #22, opened 2026-05-08
feat(embedding): add OpenAI-compatible embedder provider

ihubanov:
> Reopened from #21 — the original branch was force-pushed to clean up the commit message (Conventional Commits prefix) and add an explicit ABCI-determinism reassurance line. GitHub blocks reopens after force-push, so this is a fresh PR with the same code.

## Motivation

Lets SAGE deployments share an existing OpenAI-compatible embeddings stack (vLLM, OpenAI, LiteLLM, Text Embeddings Inference, llama.cpp server) instead of running Ollama or relying on the hash default. Useful for non-English content where multilingual embedders (e.g. gte-Qwen2-1.5B-instruct, BAAI bge-m3) noticeably outperform the hash provider.

## What changes

- New `embedding.provider` value: `openai-compatible`
- Adds `OpenAICompatibleClient` alongside the existing hash and ollama providers
- Calls `POST {base_url}/v1/embeddings` with the standard OpenAI embeddings request shape
- Configurable dimension (validated against response), optional bearer auth via `embedding.api_key`
- Optional `Named` interface so the REST `/v1/embed` info handler reports the correct provider name
- Unit tests using `httptest.NewServer` (happy path, auth header, error cases)

## ABCI determinism preserved

The embedder is only invoked from the API/web path (`api/rest`, `web/handler.go`, `cmd/sage-gui/{node,wizard}.go`). `internal/abci` stores and compares the resulting vector + hash but never calls the embedder during `Commit` / `FinalizeBlock` — so swapping providers is transparent to consensus.

## Tested

- `go test -race -count=1 ./...` clean
- `golangci-lint v2.1.6 run ./...` 0 issues
- Smoke-tested in a SAGE Personal-mode container against vLLM serving gte-Qwen2-1.5B-instruct (1536-dim, multilingual). Prose-only recall queries that returned 0 hits on the hash embedder return the expected fact on the new provider.

## Configuration example

\`\`\`yaml
embedding:
  provider: openai-compatible
  base_url: https://api.openai.com  # or vLLM, LiteLLM, your own gateway
  model: text-embedding-3-small
  api_key: sk-...
  dimension: 1536
\`\`\`

### DIFF
--- a/api/rest/embed_handler.go
+++ b/api/rest/embed_handler.go
@@ -2,6 +2,8 @@ package rest
 
 import (
 	"net/http"
+
+	"github.com/l33tdawg/sage/internal/embedding"
 )
 
 // EmbedRequest is the request body for POST /v1/embed.
@@ -46,7 +48,15 @@ func (s *Server) handleEmbedInfo(w http.ResponseWriter, r *http.Request) {
 	semantic := s.embedder.Semantic()
 	provider := "hash"
 	if semantic {
+		// Default semantic name is "ollama" for backward compat with clients
+		// that pre-date the multi-provider patch. Providers may override by
+		// implementing the optional Named interface (e.g. openai-compatible).
 		provider = "ollama"
+		if named, ok := s.embedder.(embedding.Named); ok {
+			if n := named.Name(); n != "" {
+				provider = n
+			}
+		}
 	}
 
 	// Vault-active forces semantic-only mode. Even if no embedder is
--- a/api/rest/embed_handler_test.go
+++ b/api/rest/embed_handler_test.go
@@ -1,6 +1,7 @@
 package rest
 
 import (
+	"context"
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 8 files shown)


l33tdawg:
Thanks @ihubanov — merged.

Really appreciated the rigor here:
- The ABCI-determinism note is correct (verified — `internal/abci/` has no `embedding` import; embedder is REST/web/wizard-only).
- Optional `Named` interface is a clean way to add a provider without churning the existing `Provider` contract or breaking the legacy `"ollama"` label.
- The dimension-mismatch test asserting `Ready` stays false is exactly the right behavior — silent corruption of the vector store would have been the worst outcome.
- `SAGE_EMBEDDING_*` overriding `OLLAMA_*` (applied after yaml parse) is the right precedence.

Two follow-ups I'll track separately, no action needed from you:

1. **Wizard UI surface** — `handleTestProvider` now accepts `openai-compatible`, but the wizard HTML/JS in `cmd/sage-gui/wizard.go` still only sends `ollama`/`hash`. So this provider is yaml/env-only for now. I'll wire up a wizard tile in a follow-up so non-technical operators can pick it.
2. **Ollama parity** — could implement `Named()` on the existing Ollama client too so `/v1/embed/info` reads from the method instead of the inferred fallback. Pure cleanup, not load-bearing.

Multilingual embedders (gte-Qwen2, bge-m3) on non-English content was a real gap — thanks for closing it. 🙏

PR #24, opened 2026-05-11
fix(cerebrum): status pill dispatches on embedding.provider

ihubanov:
Follow-up to #22.

Before this fix the CEREBRUM dashboard's `/v1/dashboard/health` endpoint hard-coded a probe to `http://localhost:11434/api/tags` and set `health.ollama` to `running` or `offline` regardless of which embedder was actually configured. The status pill at the top of CEREBRUM and the System Status row on the Settings → Overview tab both read that single field, so any deployment that picked `openai-compatible` (PR #22, shipped in v6.8.7) or `hash` got a permanent red "Ollama offline" indicator even when embeddings were flowing fine through the configured upstream. Purely cosmetic, but it actively misleads the operator about which dependency to chase when something does go wrong.

## Backend (`web/handler.go` `handleHealth`)

- Type-assert `h.embedder` against the optional `embedding.Named`, `embedding.Modeler`, `embedding.Pinger` interfaces (plus the `Dimension` / `Ready` / `Semantic` trio every `Provider` already has).
- Return a structured `embedder` block:

  ```json
  {
    "provider": "openai-compatible",
    "model": "Alibaba-NLP/gte-Qwen2-1.5B-instruct",
    "dimension": 1536,
    "ready": true,
    "semantic": true,
    "online": true
  }
  ```

  `online` prefers a live `Pinger` probe with a 2s context timeout; falls back to `Ready()` for providers without a `Ping`; hash is always online (no upstream).
- The legacy `ollama` string field is kept but flips to `"n/a"` when the configured provider isn't Ollama, so pre-fix dashboards stop rendering a false-positive "Ollama running" for non-Ollama deployments. Ollama deployments still see `"running"` / `"offline"` as before.

## New optional interfaces (`internal/embedding/provider.go`)

- `Modeler { Model() string }` — surfaces the configured model name in the UI; useful when an operator runs several embedding models on the same vLLM / LiteLLM endpoint.
- `Pinger { Ping(ctx) error }` — cheap liveness check separate from `Ready()`, which is a sticky has-ever-succeeded flag. For an operator-facing pill we want a real-time signal.

Both the Ollama client (`internal/embedding/ollama.go`) and the OpenAI-compatible client (`internal/embedding/openai_compatible.go`) now implement `Modeler`. Ollama already had `Ping`; the OpenAI-compatible client gets a `Ping` that issues a minimal `/v1/embeddings` call — the spec doesn't define a health path and most servers return fast on a one-token input.

## Frontend (`web/static/js/app.js`)

- New `describeEmbedder(health)` helper normalizes both the v6.8.8+ embedder block and the legacy `ollama` string into a single `{ provider, online, label, shortLabel, detail }` view so the two render sites (top health bar + Settings System Status row) stay in sync.
- Labels:

  | provider             | label |
  |----------------------|-------|
  | `ollama`             | Ollama connected/offline |
  | `openai-compatible`  | OpenAI-compatible connected/offline |
  | `hash`               | Hash provider (no semantic search) |
  | `none`               | No embedder configured |

  plus a `(model, N-dim)` suffix when the server reports them.
- Pre-v6.8.8 fallback: when only `health.ollama` is present and is `"running"` or `"offline"`, render the Ollama label. When it is `"n/a"` or absent, render "Embedder status unknown" rather than fabricating an Ollama status.

## Tests (`web/handler_health_embedder_test.go`, 6 cases)

- ollama running / offline
- openai-compatible online / offline (the bug-fix case — provider correctly reported as `openai-compatible`, legacy `ollama` field flips to `"n/a"`)
- hash (`Semantic()=false` fallback, no model suffix, always online)
- not configured (embedder block still present with `provider="none"`)

```
$ go test ./web/ -run TestHandleHealth_Embedder -v
--- PASS: TestHandleHealth_EmbedderOllamaRunning (0.18s)
--- PASS: TestHandleHealth_EmbedderOllamaOffline (0.20s)
--- PASS: TestHandleHealth_EmbedderOpenAICompatibleOnline (0.22s)
--- PASS: TestHandleHealth_EmbedderOpenAICompatibleOffline (0.20s)
--- PASS: TestHandleHealth_EmbedderHash (0.20s)
--- PASS: TestHandleHealth_EmbedderNotConfigured (0.21s)
PASS
```

`go test ./...` is green across the tree. `golangci-lint v2.1.6 run ./web/ ./internal/embedding/` reports `0 issues`.

## Local verification

Verified against a locally rebuilt `6.8.7-cerebrum-pill` image swapped into a stack pointed at `gte-Qwen2-1.5B-instruct` over the openai-compatible provider. `GET /v1/dashboard/health` returned:

```json
{
  "embedder": {
    "dimension": 1536,
    "model": "Alibaba-NLP/gte-Qwen2-1.5B-instruct",
    "online": true,
    "provider": "openai-compatible",
    "ready": false,
    "semantic": true
  },
  "ollama": "n/a",
  ...
}
```

and the pill in the browser renders `OpenAI-compatible connected (Alibaba-NLP/gte-Qwen2-1.5B-instruct, 1536-dim)` with a green dot. After verification the local stack was reverted to the upstream `:6.8.7` image.

## Scope

BFT consensus and the app-validator layer are untouched. This change is entirely below the embedder API surface and above the validator layer; the embed call path itself is unmodified.

### DIFF
--- a/internal/embedding/ollama.go
+++ b/internal/embedding/ollama.go
@@ -134,6 +134,12 @@ func (c *Client) Name() string {
 	return "ollama"
 }
 
+// Model implements embedding.Modeler so the CEREBRUM status pill can show
+// which Ollama model is currently bound (e.g. "nomic-embed-text").
+func (c *Client) Model() string {
+	return c.model
+}
+
 // Ping checks if Ollama is reachable.
 func (c *Client) Ping(ctx context.Context) error {
 	httpReq, err := http.NewRequestWithContext(ctx, "GET", c.baseURL+"/api/tags", nil)
--- a/internal/embedding/openai_compatible.go
+++ b/internal/embedding/openai_compatible.go
@@ -160,6 +160,13 @@ func (c *OpenAICompatibleClient) Name() string {
 	return "openai-compatible"
 }
 
+// Model implements embedding.Modeler so the CEREBRUM status pill can show
+// which upstream model identifier this client sends in its embed requests
+// (e.g. "Alibaba-NLP/gte-Qwen2-1.5B-instruct").
+func (c *OpenAICompatibleClient) Model() string {
+	return c.model
+}
+
 // Ping checks whether the configured endpoint is reachable. The OpenAI
 // embeddings spec doesn't define a dedicated health endpoint, so we issue
 // a minimal embed request — most servers will return a fast response or a
--- a/internal/embedding/provider.go
+++ b/internal/embedding/provider.go
@@ -22,3 +22,22 @@ type Provider interface {
 type Named interface {
 	Name() string
 }
+
+// Modeler is an optional interface a Provider can implement to expose the
+// model identifier it serves. The CEREBRUM dashboard surfaces this in the
+// embedder status pill so operators running multi-model stacks (vLLM /
+// LiteLLM / Ollama with several embedding models loaded) can confirm at a
+// glance which one SAGE actually talks to. Providers that don't implement
+// this simply don't get a model label in the UI.
+type Modeler interface {
+	Model() string
+}
+
+// Pinger is an optional interface a Provider can implement to expose a
+// cheap liveness check. The dashboard health endpoint prefers Ping over
+// Ready when present, because Ready is a sticky "has-ever-succeeded" flag
+// — useful for /v1/embed/info but unhelpful for a real-time operator pill
+// where the upstream embed server may have gone away after boot.
+type Pinger interface {
+	Ping(ctx context.Context) error
+}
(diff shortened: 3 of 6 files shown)


l33tdawg:
Merged and shipped as **v6.8.8** — tag pushed, release workflow running.

Clean follow-through on #22, much appreciated. The interface shape (`Modeler` / `Pinger` as separate optional surfaces on top of `Provider`) is exactly how I'd have factored it — the structured `embedder` block with the legacy `ollama` field flipping to `"n/a"` for non-Ollama providers is the right call for backward-compat with operators running stale dashboard JS during the upgrade window.

Six-case test matrix covering ollama running/offline × openai-compat online/offline × hash × not-configured is also more thorough than what shipped in #22, nice. Thanks for landing the loose end.

PR #25, opened 2026-05-26
fix(v8): governance + ancestor walks — three v8.0 cleanup items

ihubanov:
> Rebased onto `f3c0931` (**v8.1.0**) on 2026-05-27 — zero file overlap with
> v8.0.1 + v8.1.0; quality gate clean on the new base. v8.1.0's
> `InsertGovProposal` Postgres mirror makes the `opToString` fix below
> operationally visible on quorum deployments (the previous `unknown_4`
> string would have landed in the new `gov_proposal.operation` column).

## What

Three small fixes uncovered while reviewing the v8.0 release window
(`890dc3f..9c102b6`). All three are real bugs or asymmetries; none of
them changes consensus-path behavior.

1. `applyGovernanceProposal` now has an explicit `OpDomainReassign`
   case returning `(nil, nil)` instead of falling through to the
   default error branch.
2. `opToString` returns `"domain_reassign"` for `OpDomainReassign`
   instead of the fallback `"unknown_4"`.
3. `IsDomainOwnerOrAncestor` now skips reserved shared names in the
   ancestor walk, matching `HasAccessOrAncestor` and
   `ResolveOwningAncestor`.

This is **part 1 of a planned 4-PR sequence** from the v8.0 review:

- **PR 1 (this one)** — three cleanup items above.
- **PR 2** — `sage_v8_fork_advisory_drift_total` counter tracking
  `IsPostV8Fork` vs `postV8Fork(h)` disagreement on the FinalizeBlock
  path. Sibling to `sage_fork_branch_total` from ff04305.
- **PR 3** — `TxTypeAgentKeyRotate` + `OpAgentKeyRotate`, natural
  sibling of `OpDomainReassign` for compromised-key recovery.
- **PR 4** — `make replay-verify` harness operationalizing the
  byte-identical-to-v7.1.1 invariant.

## Why

Each fix maps to a concrete operational pain.

**B1 — `applyGovernanceProposal` missing `OpDomainReassign`.** The
switch at `internal/abci/app.go:3528` had cases for the three validator
ops then a default that returned
`fmt.Errorf("unknown governance operation: %d", ...)`. Every successful
3/4 supermajority for a domain-reassign proposal therefore tripped the
default branch, which `FinalizeBlock` logs as `"failed to apply
governance proposal"`. The on-chain effect was correct (the actual
reassignment runs in the follow-up `TxTypeDomainReassign` the admin
submits after `Status=Executed`), but operators tailing logs saw a
phantom failure on every successful proposal.

**B2 — `opToString` missing `OpDomainReassign`.** The fallback path
returned `"unknown_4"`, which landed in the off-chain
`gov_proposal.operation` mirror column for every domain-reassign
proposal. Anyone querying Postgres or wiring a dashboard on that
column saw an opaque string forever.

**D1 — `IsDomainOwnerOrAncestor` lacked the shared-domain barrier.**
`HasAccessOrAncestor` and `ResolveOwningAncestor` both skip reserved
shared names during the dotted-domain walk (see the
`SharedDomainBarrier` tests in `badger_ancestor_test.go`).
`IsDomainOwnerOrAncestor` was the asymmetric outlier. Not exploitable
today — the ABCI auto-register paths refuse shared domains, so the
ownership row should never exist — but a latent inconsistency: the
three walks disagreed on what counts as an inheritable ancestor, and
the first future code path that wrote such a row would silently
cascade ownership of `general` into every `*.general` descendant.

## How

- `internal/abci/app.go:3531` — explicit `case
  governance.OpDomainReassign: return nil, nil` with a comment
  naming why acceptance has no validator-set effect (execution lives
  in the follow-up tx).
- `internal/abci/app.go:3565` — explicit `case
  governance.OpDomainReassign: return "domain_reassign"`.
- `internal/store/badger.go:885` — `if IsSharedDomainName(ancestor)
  { continue }` inside the walk, with a comment referencing the
  matching barrier in the two sibling functions and the
  defence-in-depth rationale.

Considered alternatives:

- Routing `OpDomainReassign` through the default branch with a
  whitelist check — rejected. Explicit cases match the existing
  switch shape and make the "acceptance is intentionally a no-op"
  contract visible at the call site.
- Hardening `IsDomainOwnerOrAncestor` callers instead of the walk
  itself — rejected. The barrier already lives at the walk layer for
  the other two functions; placing the third one at a different layer
  would re-introduce the asymmetry this PR is closing.

## ABCI determinism

Consensus path untouched.

- B1 changes only a log line. The previous return was an error that
  `FinalizeBlock` logged and discarded — no proposal state, no
  validator set, no Postgres mirror row was affected by the error
  value itself.
- B2 changes only the off-chain mirror column. `opToString` is not
  read on the consensus path.
- D1 changes the result of `IsDomainOwnerOrAncestor` only in cases
  that should never produce `true` today (no shared-domain ownership
  rows exist on any live chain). No `FinalizeBlock` state transitions
  observe a different value as a result of this PR.

No `time.Now()`, no map iteration, no goroutines, no external I/O
introduced. No dependency bumps, no unrelated refactors.

## Tests

- `internal/abci/domain_reassign_test.go`
  - `TestApplyGovernanceProposal_OpDomainReassign_NoValidatorUpdate`
    — pins the `(nil, nil)` return contract.
  - `TestOpToString_OpDomainReassign` — table-driven across all four
    ops.
- `internal/store/badger_ancestor_test.go`
  - `TestIsDomainOwnerOrAncestor_SharedDomainBarrier` — `.general`
    descendant, `sage-*` descendant, direct shared-domain ownership,
    non-shared cascade sanity case.

Gates:

- `go build ./...` clean
- `go vet ./...` clean
- `go test -race -count=1 ./internal/abci/... ./internal/store/...`
  clean
- `golangci-lint v2.1.6` (CI pin) clean

Diff stat:

```
 internal/abci/app.go                   | 11 ++++++++
 internal/abci/domain_reassign_test.go  | 51 ++++++++++++++++++++++++++++++++++
 internal/store/badger.go               | 13 +++++++++
 internal/store/badger_ancestor_test.go | 42 ++++++++++++++++++++++++++++
 4 files changed, 117 insertions(+)
```


l33tdawg:
Reviewed for replay safety (the bar for anything on the consensus path) — all three changes are clean:

1. **`applyGovernanceProposal` `OpDomainReassign` case** — verified at the call site (`app.go:592`): the returned `applyErr` is log-only. `update` is nil in both the old (default→`nil, err`) and new (`nil, nil`) paths, and the `StatusExecuted` write is unconditional. So validator set, proposal status, and app hash are byte-identical old vs new — only a spurious error log line goes away. No replay divergence.
2. **`opToString` → `"domain_reassign"`** — feeds only the off-chain Postgres `gov_proposal.operation` mirror + logs, not consensus state.
3. **`IsDomainOwnerOrAncestor` shared-domain barrier** — a no-op on real chains (the triggering ownership row can't exist via the ABCI auto-register gate); pure defense-in-depth aligning the three ancestor walks. Good catch.

Additive (+117/−0), well-tested, CI green incl. Byzantine Fault Tests. Merging.

PR #27, opened 2026-05-27
fix(store): stream-hash ComputeAppHash — drop O(state) per-block alloc (#26)

### LINKED ISSUE #26: perf(store): ComputeAppHash allocates O(state-size) per block — per-block CPU grows linearly with chain height
## Symptom

`ghcr.io/l33tdawg/sage` Personal mode has per-block bookkeeping cost that grows monotonically with chain height even with zero client traffic. CometBFT block-finalize loop runs at ~1 block/sec regardless of `txs`; each block costs more than the last. Consensus stays healthy throughout (height advances, no panics, no peer storms) — it's a resource leak, not distress, but unbounded.

| Container | Image | Height | Avg CPU (one core) |
|---|---|---|---|
| fresh (~60s uptime) | `ghcr.io/l33tdawg/sage:6.8.8` | 42 | 18.98% |
| aged (~16 days uptime) | `ghcr.io/l33tdawg/sage:6.8.8` | 769,839 | **98.86%** |

Same image, zero client traffic both sides. ~5× CPU ratio = per-block work scales with state.

Confirmed unchanged on `v8.1.0` (latest): fresh-CPU `18.74%` at height ~260, identical within noise. Changelog scan v7.0 → v8.1.0 (14 releases) has zero mentions of `pruning`, `cpu`, `perf`, `leak`, `compaction`, `finalize`, `blockstore`, `statestore`, or `gc`.

Secondary signal: aged `/root/.sage` = 1.8 GB at height 769k, ~2.3 KB/block accumulation, no visible pruning.

## Root cause

`internal/store/badger.go:218` — `BadgerStore.ComputeAppHash`, called from `internal/abci/app.go:649` on **every** `FinalizeBlock`:

```go
func (s *BadgerStore) ComputeAppHash() ([]byte, error) {
    h := sha256.New()
    err := s.db.View(func(txn *badger.Txn) error {
        opts := badger.DefaultIteratorOptions
        opts.PrefetchValues = true                  // prefetches every value into iterator buffer
        it := txn.NewIterator(opts)
        // ...
        type kv struct { key, val []byte }
        var entries []kv                            // ← allocates O(state-size) per block

        for it.Rewind(); it.Valid(); it.Next() {
            // copies key + value into the slice for every key in DB
            entries = append(entries, kv{key: k, val: val})
        }
        sort.Slice(entries, ...)                    // ← redundant — Badger iter already lex-orders
        for _, e := range entries { h.Write(e.key); h.Write(e.val) }
        // ...
    })
}
```

Two compounding issues:

1. **Allocates the full DB content into a Go slice on every block.** At 1.8 GB on-disk state with values prefetched into the iterator buffer, this is ~hundreds of MB of garbage per second hitting the Go GC.
2. **Sorts the slice for "safety".** Badger's iterator already yields entries in lexicographic key order by default (LSM merge iterator). The sort is redundant and adds O(N log N) on top of an already-O(N) walk.

Combined: the GC mark/sweep cycle dominates CPU on aged chains. Per-block compute may itself fit in budget, but the GC pressure from the per-block allocation tax doesn't.

## Reproduction

Bench in `internal/store/badger_apphash_bench_test.go` (proposed by this issue):

```
goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i7-8565U @ 1.80GHz

N=1,000:    8 ms   /   473 KB   /   3,551 allocs   per ComputeAppHash call
N=10,000:  93 ms   /   5.1 MB   /  30,533 allocs   per ComputeAppHash call
N=50,000: 281 ms   /  26.2 MB   / 150,606 allocs   per ComputeAppHash call
N=100,000: 705 ms  /  52.6 MB   / 300,730 allocs   per ComputeAppHash call
```

Linear in N for time AND bytes AND allocations. At realistic aged-chain state sizes of 200k–500k live keys this is 1.4–3.5 seconds of work per block at ~1 block/sec block rate, on dedicated CPU. Production hits 99% as soon as the GC concurrent cycle starts eating wall-clock time.

## Proposed fix (Tier A, low-risk)

Stream-hash directly in iteration order. BadgerDB iterators yield keys in lexicographic order by default (`IteratorOptions.Reverse=false`, LSM merge iterator guarantee), so the explicit sort is unnecessary. No need to copy keys/values into a Go slice — `h.Write` consumes synchronously inside the iterator callback before `Next()` invalidates the underlying slice.

```go
func (s *BadgerStore) ComputeAppHash() ([]byte, error) {
    h := sha256.New()
    err := s.db.View(func(txn *badger.Txn) error {
        opts := badger.DefaultIteratorOptions
        opts.PrefetchValues = false
        it := txn.NewIterator(opts)
        defer it.Close()
        for it.Rewind(); it.Valid(); it.Next() {
            item := it.Item()
            h.Write(item.Key())
            if err := item.Value(func(v []byte) error {
                h.Write(v)
                return nil
            }); err != nil { return err }
        }
        return nil
    })
    if err != nil { return nil, err }
    return h.Sum(nil), nil
}
```

Side-by-side benchmark with the legacy implementation, identical seeded state:

```
N=1,000      Legacy:   1.69 ms   /  479 KB    /   3,563 allocs
N=1,000      Streaming: 0.98 ms /  1.78 KB   /      19 allocs

N=10,000     Legacy:   19.1 ms  / 5.1 MB     /  30,535 allocs
N=10,000     Streaming: 18.2 ms /  2.14 KB   /      21 allocs

N=100,000    Legacy:    355 ms  /  55 MB     / 300,787 allocs
N=100,000    Streaming: 203 ms  /   9 KB     /     166 allocs
```

Allocation reduction: ~6000× at N=100k. The raw ns/op speedup looks modest (~1.75×) on dedicated bench CPU, but the production win is the elimination of ~55 MB/sec garbage that triggers concurrent GC on shared production CPU.

Byte-identical hash output is pinned by `TestComputeAppHash_StreamingEqualsLegacy` across N=0, 1, 1k, 10k — both implementations produce the same SHA-256 across all populated states, so the fix is replay-safe and does not require a consensus fork.

## Optional Tier B (follow-up)

Empty blocks (txs=0) recompute the hash over identical state every second. A dirty-bit cache on `BadgerStore` — invalidated by any `Set`/`Delete`, returning the cached hash if clean — would make empty-block AppHash O(1). Bigger semantic claim, needs careful threading through every write path. Better as a separate PR after Tier A lands.

## Want to ship the fix?

Bench, regression test, and streaming implementation are ready to PR. Would prefer to confirm the fix shape with you here before opening, in case you'd rather see Tier B in the same PR, or a different approach (e.g. content-hash chain like Tendermint's `app_hash`, or moving the hash off the consensus path entirely). The bench framework is reusable for either direction.

l33tdawg:
## What

Rewrites `BadgerStore.ComputeAppHash` (`internal/store/badger.go`) to stream `h.Write(key); h.Write(val)` in BadgerDB iteration order, dropping the per-block O(state) slice copy and the redundant sort.

Fixes #26.

## Why

`ComputeAppHash` runs on **every** `FinalizeBlock` and collected every key/value into a Go slice, then re-sorted it. BadgerDB's default forward iterator already yields keys in lexicographic byte order — exactly the `string(a) < string(b)` order the hash uses — so the sort was redundant, and the slice copy made per-block allocation (and thus GC/CPU) grow linearly with chain height. @ihubanov measured ~99% of a core at height 769k vs ~19% fresh on the same image with zero traffic.

## Consensus safety

This is a **consensus no-op**: streaming in iteration order produces a **byte-identical** app hash to the previous sort-then-hash, because the orders are identical. No fork, no fork gate, no replay divergence. `TestComputeAppHash_ByteIdenticalToSortedConcat` pins this by reproducing the old sort-then-hash inline and asserting equality; `TestComputeAppHash_OrderIndependent` pins write-order independence.

## Numbers (`-benchmem`)

| N | before (issue) | after |
|---|---|---|
| 1,000 | 473 KB / 3,551 allocs | 45 KB / 1,717 allocs |
| 10,000 | 5.1 MB / 30,533 allocs | 198 KB / 10,833 allocs |
| 50,000 | 26.2 MB / 150,606 allocs | 829 KB / 50,912 allocs |

Local `go test ./internal/store/ ./internal/abci/` green.

Diagnosis and Tier-A fix by @ihubanov in #26.

### DIFF
--- a/internal/store/badger.go
+++ b/internal/store/badger.go
@@ -224,36 +224,25 @@ func (s *BadgerStore) ComputeAppHash() ([]byte, error) {
 		it := txn.NewIterator(opts)
 		defer it.Close()
 
-		// Collect all keys for sorting (BadgerDB iterates in sorted order by default,
-		// but we make it explicit for safety)
-		type kv struct {
-			key, val []byte
-		}
-		var entries []kv
-
+		// BadgerDB's default forward iterator yields keys in lexicographic
+		// byte order — exactly the order the app hash requires (it matches a
+		// `string(a) < string(b)` sort). So we stream each key+value straight
+		// into the digest in iteration order. This is byte-identical to the
+		// previous collect-into-slice-then-sort approach (same input → same
+		// hash, so no consensus change), but avoids allocating the entire DB
+		// into a Go slice on every FinalizeBlock — that per-block allocation
+		// made GC pressure (and thus CPU) grow linearly with chain height.
+		// See issue #26. h.Write consumes the key/value synchronously, so the
+		// iterator's borrowed slices stay valid for the duration of the call.
 		for it.Rewind(); it.Valid(); it.Next() {
 			item := it.Item()
-			k := make([]byte, len(item.Key()))
-			copy(k, item.Key())
-			valErr := item.Value(func(v []byte) error {
-				val := make([]byte, len(v))
-				copy(val, v)
-				entries = append(entries, kv{key: k, val: val})
+			h.Write(item.Key())
+			if valErr := item.Value(func(v []byte) error {
+				h.Write(v)
 				return nil
-			})
-			if valErr != nil {
+			}); valErr != nil {
 				return valErr
 			}
-		}
-
-		// Sort by key for determinism
-		sort.Slice(entries, func(i, j int) bool {
-			return string(entries[i].key) < string(entries[j].key)
-		})
-
-		for _, e := range entries {
-			h.Write(e.key)
-			h.Write(e.val)
 		}
 
 		return nil
--- a/internal/store/badger_apphash_test.go
+++ b/internal/store/badger_apphash_test.go
@@ -1,0 +1,136 @@
+package store
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"sort"
+	"testing"
+
+	badger "github.com/dgraph-io/badger/v4"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// referenceAppHash reproduces the pre-#26 algorithm: collect every key/value
+// into a slice, sort by key, then hash key||value in sorted order. The
+// streaming ComputeAppHash must be byte-identical to this — that equality is
+// what guarantees issue #26's perf fix did NOT change the app hash (i.e. it is
+// a consensus no-op, not a fork). If this ever diverges, the fix changed the
+// committed app hash and would fork existing chains.
+func referenceAppHash(t *testing.T, s *BadgerStore) []byte {
+	t.Helper()
+	type kv struct{ k, v []byte }
+	var entries []kv
+	require.NoError(t, s.db.View(func(txn *badger.Txn) error {
+		opts := badger.DefaultIteratorOptions
+		opts.PrefetchValues = true
+		it := txn.NewIterator(opts)
+		defer it.Close()
+		for it.Rewind(); it.Valid(); it.Next() {
+			item := it.Item()
+			k := append([]byte(nil), item.Key()...)
+			var v []byte
+			require.NoError(t, item.Value(func(val []byte) error {
+				v = append([]byte(nil), val...)
+				return nil
+			}))
+			entries = append(entries, kv{k, v})
+		}
+		return nil
+	}))
+	sort.Slice(entries, func(i, j int) bool { return string(entries[i].k) < string(entries[j].k) })
+	h := sha256.New()
+	for _, e := range entries {
+		h.Write(e.k)
+		h.Write(e.v)
+	}
+	return h.Sum(nil)
+}
+
+// TestComputeAppHash_ByteIdenticalToSortedConcat is the consensus-safety pin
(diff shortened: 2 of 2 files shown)


ihubanov:
Two small offers (no urgency — happy if either gets passed on):

**1. `opts.PrefetchValues = false`**

The streaming loop only needs one value at a time, so Badger's prefetch buffer ends up holding values that get immediately consumed and discarded. Setting it to `false` lets Badger read each value lazily on the `.Value()` call. Local bench against the same streaming approach: dropping prefetch took per-call allocation from ~45 KB / ~1,700 allocs at N=1k down to ~2 KB / ~20 allocs (similarly at higher N). On-the-wire hash is identical either way.

**2. Bench harness — `BenchmarkComputeAppHash_GrowsLinearlyWithState`**

A parameterized bench drop-in for `internal/store/` that drives the function across N=1k/10k/50k/100k with synthetic state. Useful as a CI regression lock to pin per-block cost flat as state grows. Happy to open a follow-up PR with just that file if you'd take it.

Either way, the diagnosis-credit means a lot — thanks for the prompt fix.

PR #29, opened 2026-06-02
fix(rest): handleEmbed — report the actual embedding model, not a hardcoded literal

ihubanov:
## What

`POST /v1/embed` now reports the model that actually produced the embedding instead of a hardcoded `"nomic-embed-text"`. A node running the `openai-compatible` embedder previously returned `model: "nomic-embed-text"` for vectors produced by a different model (e.g. `Alibaba-NLP/gte-Qwen2-1.5B-instruct`).

## Why

`handleEmbed` (`api/rest/embed_handler.go`) wrote `Model: "nomic-embed-text"` literally into every `EmbedResponse`, regardless of the configured provider. `api/openapi.yaml:1234` documents `EmbedResponse.model` as the model name with no caveat, so a stale literal contradicts the spec.

The `embedding.Modeler` optional interface (`internal/embedding/provider.go:32`, `Model() string`) already exists for exactly this. Both real providers implement it — `OpenAICompatibleClient.Model()` (`internal/embedding/openai_compatible.go:166`) returns its configured model, Ollama `Client.Model()` (`internal/embedding/ollama.go:139`) returns its bound model. The sibling `handleEmbedInfo` in the same file already feature-detects the optional `Named` interface to report the real `provider` (`api/rest/embed_handler.go:55`); `handleEmbed` was never given the parallel treatment for `model`.

This completes wiring that landed in PR #22 (`feat(embedding): add OpenAI-compatible embedder provider`) — that PR added the provider and the `Modeler` interface, but `/v1/embed`'s response stayed hardcoded. It's the same misreport class as PR #24 (`fix(cerebrum): status pill dispatches on embedding.provider`): a surface lying about the active provider/model, in a handler #24 didn't touch.

## How

- after computing the embedding, feature-detect `embedding.Modeler` and use `Model()` when present and non-empty
- fall back to the legacy `"nomic-embed-text"` literal only for providers that don't implement `Modeler` (the hash provider), preserving prior behavior on that path
- mirrors `handleEmbedInfo`'s `Named` feature-detect pattern exactly

## ABCI determinism

No consensus-path code touched. This is a read/util REST handler that formats a response; no tx, no `internal/abci`, no store write, no AppHash contribution.

## Tests

- `TestHandleEmbed_ModelerReportsActualModel` — a `Modeler`-implementing embedder makes `/v1/embed` report its actual model (`Alibaba-NLP/gte-Qwen2-1.5B-instruct`), not the default; mirrors `TestHandleEmbedInfo_NamedProviderOverridesOllama`
- `TestHandleEmbed_FallsBackForNonModeler` — the hash provider (no `Modeler`) still returns the legacy `nomic-embed-text` default; pins no-regression on that path
- `go build ./...` clean
- `go test -race -count=1 ./api/rest/...` clean
- `golangci-lint v2.12.2` (current CI pin) clean

### DIFF
--- a/api/rest/embed_handler.go
+++ b/api/rest/embed_handler.go
@@ -103,9 +103,22 @@ func (s *Server) handleEmbed(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Report the model the embedding was actually produced with. Providers
+	// expose it via the optional embedding.Modeler interface (openai-compatible
+	// returns its configured model, Ollama returns its bound model). Mirrors
+	// handleEmbedInfo's feature-detect of the Named interface for `provider`.
+	// Falls back to the legacy default for providers that don't implement
+	// Modeler (e.g. the hash provider), preserving prior behavior there.
+	model := "nomic-embed-text"
+	if m, ok := s.embedder.(embedding.Modeler); ok {
+		if name := m.Model(); name != "" {
+			model = name
+		}
+	}
+
 	writeJSON(w, http.StatusOK, EmbedResponse{
 		Embedding: emb,
-		Model:     "nomic-embed-text",
+		Model:     model,
 		Dimension: len(emb),
 	})
 }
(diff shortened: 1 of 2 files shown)


l33tdawg:
Merged and shipped in **v8.8.1**. Clean fix - verified the `embedding.Modeler` feature-detect mirrors the existing `Named` pattern in `handleEmbedInfo` exactly, build + race tests + both new tests green locally. Appreciate the thorough writeup and the no-regression test on the hash-provider fallback path. Thanks for the contribution, @ihubanov 🙏

PR #30, opened 2026-06-02
fix(sdk): MemoryRecord — read back the provider provenance tag the server emits

ihubanov:
## What

Adds an Optional `provider` field to the Python SDK's `MemoryRecord`
(`sdk/python/src/sage_sdk/models.py`) so the SDK reads back the provenance tag
the server already emits on every memory record.

## Why

The server carries `provider` on every record: the Go `memory.MemoryRecord` has
`Provider string` with `json:"provider,omitempty"`
(`internal/memory/model.go:48`), captured at submit time from the submitter's
request and persisted onto the committed record. It's the submitter's
provenance tag (e.g. `"claude-code"`, `"chatgpt"`), not the embedding model
name.

The SDK model had no matching field, so it dropped `provider` when parsing
responses from `list_memories()`, `get_memory()`, `query()`, and `hybrid()`.
The asymmetry: `list_memories()` and `query()` already accept `provider` as a
filter parameter (`sdk/python/src/sage_sdk/client.py:260`, `:292`, `:358`,
`:464`) — a caller could filter results by provider but never see which
provider a returned record carried. Filter-but-not-read.

Same class as PR #24 (Optional SDK model fields for server response fields the
SDK couldn't read back) and PR #29 (`handleEmbed` reporting the actual
embedding model). v8.6.0 added Optional PoE fields to `AgentProfile` the same
way. This is the established pattern for SDK round-trip gaps.

## How

- one additive Optional field: `provider: str | None = None`. Additive +
  Optional means older servers that omit it (`omitempty`) parse cleanly to
  `None` — forward/back compatible.
- the model is shared by the sync and async clients, so both round-trip
  `provider` with a single change.
- `MemorySubmitRequest` is unchanged — this is read-path only; the field is not
  newly settable from the SDK.
- adds `test_memory_record_parses_provider` and
  `test_memory_record_tolerates_missing_provider`, mirroring the existing
  present/absent PoE-signal test pair.

## Tests

- `pip install -e ".[dev]" && pytest` → 100 passed (12 model tests including
  the 2 new ones, plus all client/async/rbac/federation/tls tests). Matches the
  CI `Python SDK Tests` job.
- SDK-only, no server or consensus code touched; no version bump (left to the
  maintainer's release flow, as PR #29 was released as v8.8.1).

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,13 +57,68 @@ Add agents, configure domain-level read/write permissions, manage clearance leve
 
 ---
 
-## What's New in v8.8.1
+## What's New in v9.2.0
+
+**Corroboration integrity and an on-chain author field.** v9.2.0 activates a new independent `app-v10` consensus fork that makes corroboration a trustworthy multi-agent signal, and exposes corroboration on the MCP surface. Like every SAGE fork it is replay-safe by construction: no existing chain has activated `app-v10`, so every historical block replays byte-identically, and the new rules apply only after an operator governance-activates the fork.
+
+- **Corroboration is now guarded in consensus.** Corroboration is the signal that moves a memory from attributed toward consensus, so it must come from independent agents. Post-`app-v10`, `FinalizeBlock` rejects a self-corroboration (an agent backing its own memory), a duplicate corroboration (the same agent backing a memory twice), and a corroboration of a memory that was never submitted. Previously none of these was checked, so a single agent could inflate a memory's corroboration weight. The checks read only on-chain state, so every validator reaches the same verdict.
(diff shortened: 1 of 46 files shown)


l33tdawg:
Thanks @ihubanov — clean and exactly the established pattern (PR #24/#29, the v8.6.0 `AgentProfile` Optional fields). Verified the round-trip: `query()`/`list_memories()` accepting `provider` as a filter while the model dropped it on read was a real filter-but-not-read asymmetry, and the additive-Optional shape keeps it forward/back compatible. Tests + all 8 checks green.

Merging now. It'll ship to PyPI in the next SDK release (v9.2.2). 🙏

l33tdawg:
Correction to my note above, @ihubanov — this didn't actually need a new release. The identical change is **already in `main`** (commit `30e05d0`) and has been shipping on PyPI since **v9.2.0** — `pip install "sage-agent-sdk>=9.2.0"` already has `MemoryRecord.provider`.

This PR's branch was cut from an older `main` (before that commit landed), so the rebase-merge resolved to a no-op against the already-merged change — which is why it closed without adding a duplicate. Net: your fix is live now, nothing further needed. Thanks again, and sorry for the v9.2.2 misdirection. 🙏

PR #42, opened 2026-06-12
fix(sdk): GovProposal — read back the created_at the server emits

ihubanov:
## What

Adds an additive Optional `created_at` field to the Python SDK `GovProposal`
model so it reads back the timestamp the server already emits on both the
governance list and detail endpoints.

## Why

The server always produces `created_at`:

- `governance_proposals.created_at` is `TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ','now'))` — populated on every insert.
- `ListGovProposals` and `GetGovProposal` both `SELECT COALESCE(created_at,'')` and scan into `GovProposal.CreatedAt` (`internal/store/sqlite.go:3736`, `json:"created_at,omitempty"`).
- `web/governance_handler.go` serializes that struct straight to the SDK on the list and detail dashboard endpoints.

The SDK `GovProposal` model ended at `reason` with no `created_at`, so a caller
using `client.governance_proposals()` / `governance_proposal_detail()` got the
timestamp silently discarded and could never see when a proposal was raised.
Same write-but-not-read asymmetry already fixed for `linked_memories` (#39) and
`provider` (#30).

## How

- `models.py`: one additive Optional field, `created_at: datetime | None = None` (`datetime` was already imported). Older servers omitting it default to `None` — back- and forward-compatible.
- `tests/test_models.py`: `test_gov_proposal_parses_created_at` (RFC3339 string → `datetime`) and `test_gov_proposal_tolerates_missing_created_at` (absent → `None`), mirroring the existing provider / linked_memories round-trip tests.

## ABCI determinism

Consensus path untouched — SDK-only change, no edits under `internal/abci/`,
`internal/store/` write paths, `internal/tx/`, or `internal/governance/`.

## Tests

16/16 pytest pass, including the two new `GovProposal` round-trip cases. Full CI green on the fork branch (Go Analyze, Lint, Test, Byzantine Fault Tests, Docker Build, CodeQL, Python SDK Tests).

### DIFF
--- a/sdk/python/src/sage_sdk/models.py
+++ b/sdk/python/src/sage_sdk/models.py
@@ -462,6 +462,13 @@ class GovProposal(BaseModel):
     expiry_height: int
     executed_height: int | None = None
     reason: str | None = None
+    # Wall-clock creation time the server stamps on every proposal row
+    # (governance_proposals.created_at, NOT NULL DEFAULT RFC3339 — always
+    # populated) and emits as `created_at` on BOTH the list and detail
+    # responses (json:"created_at,omitempty"). The model dropped it on read, so
+    # a caller listing/inspecting proposals could never see when each was
+    # raised. Additive Optional: older servers omit it, so it defaults to None.
+    created_at: datetime | None = None
 
 
 class GovVote(BaseModel):
--- a/sdk/python/tests/test_models.py
+++ b/sdk/python/tests/test_models.py
@@ -92,6 +92,42 @@ def test_submit_request_valid():
     assert req.content == "Test memory content"
 
 
+def test_gov_proposal_parses_created_at():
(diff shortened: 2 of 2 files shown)


l33tdawg:
Shipped in **v10.5.3** 🎉

Thanks! Clean additive read-back — same write-but-not-read asymmetry you (and the team) already closed for `linked_memories` (#39) and `provider` (#30). `GovProposal.created_at` now round-trips, and the Optional default keeps it back/forward compatible with older servers. The two round-trip tests made it trivial to verify.

Live now on PyPI: `pip install -U sage-agent-sdk==10.5.3` — full release at https://github.com/l33tdawg/sage/releases/tag/v10.5.3

PR #43, opened 2026-06-12
fix(rest): add-member handlers — honor explicit clearance:0 (PUBLIC)

ihubanov:
## What

The org and dept add-member REST endpoints silently escalate an explicitly-requested PUBLIC clearance (`clearance: 0`) to INTERNAL (`1`). After this fix, an explicit `0` is carried verbatim into the broadcast tx; an omitted `clearance` field still defaults to INTERNAL.

## Why

`clearance: 0` is `ClearancePublic` — a real, named clearance level ("Readable by any federated org", `internal/tx/types.go:90`), not a sentinel. It gates reads: a member's clearance is compared `clearance >= memoryClassification` (`internal/store/badger.go:2812`), so a clearance-0 member should read only PUBLIC memories.

`handleOrgAddMember` (`api/rest/org_handler.go`) and `handleDeptAddMember` (`api/rest/dept_handler.go`) declared `Clearance int` and ran `if req.Clearance == 0 { req.Clearance = 1 /* Default to INTERNAL */ }`. A bare int cannot distinguish an explicit `0` from an omitted field, so an explicit `clearance: 0` was coerced to `1` — granting the member INTERNAL read access they did not request. Because the SDK's `DeptAddMemberRequest`/`OrgAddMember` default `clearance: int = 1`, the only way to *ask* for PUBLIC is to send `0`, which the server then bumped back to `1`. The PUBLIC-clearance member was unreachable end to end.

This is the identical bug class already fixed on the agent-permission endpoint — "Bug 2" in the v6.8.4 hotfix bundle, documented at `api/rest/agent_handler.go:281` ("would silently demote the agent to clearance=1"). That path was fixed by making the request field a pointer (`Clearance *int`, `agent_handler.go:245`) so absent and explicit-0 are distinguishable, with a regression test (`TestSetPermission_PartialUpdate_PreservesClearance`, `api/rest/permission_handler_test.go`). This applies the same established pattern to the two add-member handlers that were left unpatched.

## How

- `OrgAddMemberReq`/`DeptAddMemberReq`: `Clearance` is now `*int`.
- Replaces `if req.Clearance == 0 { req.Clearance = 1 }` with `clearance := 1; if req.Clearance != nil { clearance = *req.Clearance }` — nil (omitted) keeps the safe INTERNAL default, explicit `0` is honored verbatim into the broadcast tx.
- Mirrors the pointer technique already in `agent_handler.go`; no new mechanism introduced.

## ABCI determinism

No consensus path is touched. This changes the REST request shape and the REST default-mapping only. The on-chain tx already carries a concrete `ClearanceLevel`; the tx encoding, the `internal/tx` types, and the `FinalizeBlock` path are unchanged. Replay is byte-identical.

## Tests

- New `api/rest/addmember_clearance_test.go`: 4 tests (org + dept × explicit-0-not-escalated + omitted-defaults-internal) that decode the captured broadcast tx and assert the `Clearance` carried, mirroring the Bug 2 test style.
- `go test ./api/rest/`, `go vet`, and golangci-lint clean. Full CI green on the fork branch (Go/JS Analyze, Lint, Test, Byzantine Fault Tests, Docker Build, CodeQL, Python SDK Tests).

### DIFF
--- a/api/rest/addmember_clearance_test.go
+++ b/api/rest/addmember_clearance_test.go
@@ -1,0 +1,118 @@
+package rest
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/l33tdawg/sage/internal/tx"
+)
+
+// These guard the org/dept add-member endpoints against the clearance-0
+// escalation: an explicit clearance of 0 (ClearancePublic — a valid level that
+// gates reads, internal/tx/types.go) must be carried verbatim into the
+// broadcast tx, not silently coerced to INTERNAL (1). Only an OMITTED clearance
+// falls back to the safe INTERNAL default. Same class as the agent-permission
+// "Bug 2" fix; here a bare int couldn't tell "0" from "absent".
+
+// newAddMemberCometMock returns a CometBFT stub that records the broadcast tx
+// hex into *captured and acks Code 0.
+func newAddMemberCometMock(t *testing.T, captured *string) *httptest.Server {
+	t.Helper()
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		*captured = r.URL.Query().Get("tx")
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(map[string]interface{}{
+			"result": map[string]interface{}{"code": 0, "hash": "ADDMEMBER_TX"},
+		})
+	}))
+}
+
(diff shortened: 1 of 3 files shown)


l33tdawg:
Shipped in **v10.5.3** 🎉

Thanks for this — the bare-`int` vs `*int` catch was exactly right, and lining it up with the established `agent_handler.go` Bug 2 pattern made it an easy merge. Explicit `clearance: 0` (PUBLIC) now rides verbatim into the broadcast tx, omitted still defaults to INTERNAL, and PUBLIC members are reachable end to end again. Appreciate the thorough writeup, the `file:line` citations, and the four broadcast-tx-decoding tests.

Out now:
- PyPI: `pip install -U sage-agent-sdk==10.5.3`
- Docker: `ghcr.io/l33tdawg/sage:10.5.3`
- Desktop installers + checksums: https://github.com/l33tdawg/sage/releases/tag/v10.5.3

PR #47, opened 2026-06-18
fix(mcp): sage_link missing from never-block-for-turn allowlist

ihubanov:
## Problem
`sage_link` writes were silently dropped under turn discipline. The turn-checkpoint mechanism (`shouldBlockForTurn`) blocks tool calls that arrive after a turn's write budget is spent, but the never-block allowlist omitted `sage_link` — so batches of link requests committed only a fraction of their rows (observed: 13 requests → 0 writes in one batch).

## Fix
Add `sage_link` to the never-block switch in `shouldBlockForTurn`, alongside the other relationship/maintenance tools that must always complete. Relationship edges are low-cost metadata writes; dropping them silently corrupts the connectome without surfacing an error.

## Verification
Clean-room harness: pre-fix, batched `sage_link` calls dropped writes non-deterministically; post-fix, all link requests persist (verified on a fresh node).

1 line changed in `internal/mcp/server.go`.

### DIFF
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -331,7 +331,7 @@ func (s *Server) shouldBlockForTurn(toolName string) bool {
 	// Never block SAGE tools themselves.
 	switch toolName {
 	case "sage_turn", "sage_inception", "sage_red_pill", "sage_reflect", "sage_recall",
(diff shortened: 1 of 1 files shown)


l33tdawg:
Shipped in **v10.8.0** as `fe1bd20` (cherry-picked verbatim). Thanks @ihubanov 🙏

PR #48, opened 2026-06-18
fix(web): clear SSE write deadline so heartbeats survive the server WriteTimeout

ihubanov:
## Problem
Dashboard SSE connections were being severed every `WriteTimeout` seconds, producing a connect/disconnect storm in the chain activity log and `ERR_INCOMPLETE_CHUNKED_ENCODING` in the browser.

The heartbeat ticker was meant to keep the stream alive, but `http.Server.WriteTimeout` is an **absolute** per-connection deadline — it is set when the connection is accepted and is *not* reset by subsequent writes. Heartbeats write into a connection that the server has already scheduled to guillotine.

## Fix
Clear the write deadline for the lifetime of the SSE handler via `http.NewResponseController(w).SetWriteDeadline(time.Time{})`. This opts the long-lived stream out of the global `WriteTimeout` without weakening it for normal (short) request handlers. Heartbeats now do their job.

Also corrected the heartbeat comment to describe the actual mechanism.

## Verification
Reproduced the disconnect cycle against a server with a short `WriteTimeout`; after the change the SSE stream stays open indefinitely and heartbeats flush cleanly.

`web/sse.go` only.

### DIFF
--- a/web/sse.go
+++ b/web/sse.go
@@ -100,6 +100,16 @@ func (b *SSEBroadcaster) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Cache-Control", "no-cache")
 	w.Header().Set("Connection", "keep-alive")
 
+	// SSE is a long-lived stream, but the http.Server has WriteTimeout set
+	// (15s). That timeout is an ABSOLUTE per-connection write deadline — the
+	// heartbeat below does NOT reset it — so without clearing it the server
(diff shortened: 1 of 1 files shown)


l33tdawg:
Shipped in **v10.8.0** as `f047e78` (cherry-picked verbatim). Confirmed the premise — the HTTP server sets `WriteTimeout: 15s`, so clearing the per-connection write deadline is the right fix; heartbeats alone never reset it. Thanks @ihubanov 🙏

PR #49, opened 2026-06-18
feat(cerebrum): optional 3D MRI memory-brain view (2D/MRI toggle)

ihubanov:
## What

Adds an optional **3D "MRI" memory-brain** view to the cerebrum dashboard, alongside the existing 2D force graph. A toggle in the top-right switches between them; the 2D view remains the default and is untouched. The 3D view maps the memory store onto the complementary-learning-systems analogy already documented in `SAGE_AGI_BRAIN_ANALOGY.md`:

- **domain** → azimuthal lobe (clustered like cortical regions)
- **corroboration_count** → radial depth (hippocampal center → cortical surface)
- **confidence** → node opacity (memory decay)
- **challenged / deprecated** → greyed out (synaptic pruning)
- **sage_link type** → edge color + particle flow (the connectome)

Layout positions are deterministic and pinned (the force simulation is disabled), so the brain is stable across reloads and live SSE updates land in the right lobe instead of drifting.

## Why

Memory at scale is hard to reason about as a flat list. The spatial layout makes corroboration depth, domain clustering, and the link graph legible at a glance — and it's a natural fit for the brain analogy the project already leans on.

## Security / privacy (by design)

- **Edge leak guard:** a typed edge is emitted only when *both* endpoints are in the RBAC-visible set. No edge can reveal the existence of a memory the caller can't already see.
- **Operator-only aggregates:** the true total and per-domain counts are gated behind `seeAll`; non-operators get only their visible slice.
- **No content/embeddings leak:** node payloads carry truncated content only; embeddings and full bodies never reach the client.
- **Bounded render:** node count is capped (`SAGE_GRAPH_MAX_NODES`, default 500) with stratified importance sampling per domain, so a large store can't be used to wedge a client browser. Drill-down (`?domain=`) re-queries within a single lobe.

## Dependencies & licensing — your call

Two things here are worth flagging explicitly so you can decide how you want to carry them:

1. **Front-end libraries via CDN.** The view loads `three` and `3d-force-graph` from `esm.sh` through an importmap (`?external=three` so a single Three instance is shared). No new vendored JS, no build step. If you'd rather vendor or self-host these, that's a drop-in change.

2. **The anatomical brain mesh is separately licensed.** `web/static/assets/brain.obj` is a decimated pial surface derived from Brainder's "Brain for Blender" (Anderson M. Winkler, brainder.org), which is **CC-BY-SA 3.0** — *not* Apache-2.0 like the rest of the repo. It's attributed in `web/static/assets/README.md`, and being a separately-licensed asset it doesn't affect the repo's Apache-2.0 license. But CC-BY-SA is copyleft, so if you'd prefer not to carry it at all:

   **There's a license-free fallback built in.** The renderer generates a procedural brain hull (`makeBrainGeometry()`) and only swaps in `brain.obj` *if the file is present*. Delete `web/static/assets/brain.obj` and the view falls back to the procedural brain automatically — **zero code change, no CC-BY-SA asset, fully Apache-clean.** It's not as pretty as the real mesh, but it keeps the tree license-free if that's the call you want to make.

So: ship the real mesh (attributed) for the nicer look, or drop the one file for the license-clean procedural brain. Both work today; I've left the decision to you.

## Notes

- 2D view and all existing endpoints are unchanged; this is purely additive behind a toggle.
- `handleGraph` gains `corroboration_count`, `status=all`, operator-only `total` + `domain_counts`, and `?domain=` drill-down.
- Verified rendering headless and against a live node with a continuous organic memory feed (live SSE growth lands in-lobe; typed links animate).


l33tdawg:
Shipped in **v10.8.0** as `28aca1d` — thank you @ihubanov, fantastic feature 🧠

Three deltas from the PR as merged, for your awareness:

1. **Dropped `brain.obj`.** It's CC BY-SA 3.0, and we keep the repo cleanly Apache-2.0, so the MRI view now defaults to the procedural `makeBrainGeometry()` hull. A CC0 `.obj` can still be dropped into `web/static/assets/` to override — the loader now validates it's a real mesh first, so a missing file can't blank the hull via the static server's SPA-200 fallback.
2. **2D canvas stays the default** brain view; 3D MRI is the opt-in toggle.
3. **No synthetic fallback** — an empty / zero-memory node renders a blank brain instead of placeholder data.

The RBAC-safe typed-edge emission (only when both endpoints are visible) was 👌. Thanks again!

PR #50, opened 2026-06-23
feat(node): make CometBFT RPC/P2P ports configurable (SAGE_CMT_*)

ihubanov:
## What

Add two env knobs — `SAGE_CMT_RPC_ADDR` and `SAGE_CMT_P2P_ADDR` — to `sage-gui serve` so the CometBFT RPC (26657) and P2P (26656) listen ports are configurable instead of hardcoded. Defaults preserve today's ports exactly; with no env set the behavior is byte-for-byte identical to current.

## Why

`sage-gui serve` hardcoded the CometBFT RPC and P2P ports, so a second node can't run alongside an existing one on the same host even with a distinct `SAGE_HOME` and `REST_ADDR` — the two collide on 26657/26656 and the second `serve` fails with "address already in use". REST was already configurable via `REST_ADDR`; the CometBFT transport was the missing half. These two knobs close that gap and make personal-node coexistence on one host possible.

## How

Two env vars, both backward-compatible (unset → historical defaults):

- `SAGE_CMT_RPC_ADDR` default `tcp://127.0.0.1:26657` — RPC listen address, and the source the tx-broadcast client target is derived from
- `SAGE_CMT_P2P_ADDR` default `tcp://127.0.0.1:26656` — P2P listen (personal mode)

Implementation in `cmd/sage-gui/node.go`:

- All five hardcoded sites route through three package-level helpers (`cmtRPCAddr`, `cmtRPCClientURL`, `cmtP2PAddr`): RPC listen (`node.go:333`), personal P2P listen (`node.go:355`), quorum P2P default (`node.go:341`), tx-broadcast client URL (`node.go:435`), and the persisted `config.toml` template's rpc laddr (`node.go:922`).
- The tx-broadcast client URL is **derived** from the RPC listen address (`tcp://` → `http://`; a `0.0.0.0` wildcard listen host is dialed as `127.0.0.1`) rather than re-hardcoded, so the client can never drift from the server it's broadcasting to.
- The in-process node uses the programmatically-built `cometCfg`; the `node.go:333` override is authoritative. The on-disk `config.toml` rpc laddr is not re-read by the in-process node, so updating the template (`node.go:922`) is cosmetic-but-consistent, not functional — stated here so reviewers don't have to wonder which value wins.
- **Quorum mode is unchanged.** The quorum P2P fallback (`node.go:341`) keeps its historical `tcp://0.0.0.0:26656` default when the env is unset — the helper takes a per-mode fallback — so multi-host quorum still binds the wildcard. Only the personal-mode default is loopback.

## ABCI determinism

Consensus path untouched. The CometBFT RPC/P2P ports are node-local transport endpoints, not part of ABCI state or block production; nothing on the FinalizeBlock path reads them. Determinism-neutral.

## Tests

- `cmd/sage-gui/cmtaddr_test.go` covers default-preservation, env override, and the `tcp://`→`http://` + `0.0.0.0`→`127.0.0.1` client-URL derivation.
- `go test -race ./...` clean; `golangci-lint` clean.

Manual 2-node coexistence check:

1. **Default unchanged** — `sage-gui serve` with no new env binds 26657/26656 as before.
2. **Coexistence** — node A on defaults (`REST_ADDR=127.0.0.1:8080`); node B with `SAGE_HOME=~/.sage-second REST_ADDR=127.0.0.1:8099 SAGE_CMT_RPC_ADDR=tcp://127.0.0.1:36657 SAGE_CMT_P2P_ADDR=tcp://127.0.0.1:36656`. Both come up with no "address already in use", and a memory write/read round-trips on each REST port — confirming each node's tx broadcast targets its own RPC.
3. **Restart** — an already-initialized second `SAGE_HOME` restarted with the env set picks up the new ports.

## Out of scope

`internal/orchestrator/bundle.go:42` also hardcodes 26656 for multi-node *bundles*. That's a separate path from `serve` and a possible future follow-up; not addressed here.

The two new vars are documented in `docs/reference/environment-variables.md` (Server & networking table).

### DIFF
--- a/cmd/sage-gui/cmtaddr_test.go
+++ b/cmd/sage-gui/cmtaddr_test.go
@@ -1,0 +1,55 @@
+package main
+
+import "testing"
+
+// TestCmtRPCAddr pins the SAGE_CMT_RPC_ADDR override and its default. The
+// default must stay the historical tcp://127.0.0.1:26657 so existing single-node
+// deployments are byte-for-byte unchanged; a set value wins verbatim so a second
+// personal node can be moved off 26657 to coexist on one host.
+func TestCmtRPCAddr(t *testing.T) {
+	t.Setenv("SAGE_CMT_RPC_ADDR", "")
+	if got := cmtRPCAddr(); got != "tcp://127.0.0.1:26657" {
+		t.Errorf("default cmtRPCAddr() = %q, want tcp://127.0.0.1:26657", got)
+	}
+	t.Setenv("SAGE_CMT_RPC_ADDR", "tcp://127.0.0.1:36657")
+	if got := cmtRPCAddr(); got != "tcp://127.0.0.1:36657" {
+		t.Errorf("override cmtRPCAddr() = %q, want tcp://127.0.0.1:36657", got)
+	}
+}
+
+// TestCmtRPCClientURL pins the listen-addr → dial-URL derivation: tcp:// becomes
+// http://, and a 0.0.0.0 wildcard listen host is dialed as 127.0.0.1 (you can't
+// reliably connect to the wildcard).
+func TestCmtRPCClientURL(t *testing.T) {
+	cases := []struct {
+		env  string
+		want string
+	}{
+		{"", "http://127.0.0.1:26657"},                      // default
+		{"tcp://127.0.0.1:36657", "http://127.0.0.1:36657"}, // moved port
+		{"tcp://0.0.0.0:26657", "http://127.0.0.1:26657"},   // wildcard → loopback dial
+	}
+	for _, c := range cases {
+		t.Setenv("SAGE_CMT_RPC_ADDR", c.env)
+		if got := cmtRPCClientURL(); got != c.want {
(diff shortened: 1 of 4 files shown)


l33tdawg:
Thanks @ihubanov — merged. 🙏 Clean, well-scoped change: defaults preserved byte-for-byte, the tx-broadcast URL derived from the listen addr (so client/server can't drift), and the quorum wildcard fallback kept. Nice work.

I folded in one follow-up before merge: `web/handler.go`'s health panel (`/status` + `/net_info`) was still dialing the hardcoded `127.0.0.1:26657`, ignoring the `DashboardHandler.CometBFTRPC` field every other RPC consumer already uses — so a node moved off 26657 would've reported another node's chain state in its UI. Now it dials the configured endpoint.

And to take it the rest of the way to full coverage, #51 closes the remaining sibling sites you flagged as out-of-scope: `sage-cli status`, the `sage-gui upgrade` RPC default, and the orchestrator bundle's `p2p_addr` (it was hardcoded non-empty, which shadowed the `SAGE_CMT_P2P_ADDR` override). Appreciate the contribution! 🚀

PR #73, opened 2026-07-14
fix(11.8.2): H-2 residual hygiene — CheckTx future-cap + EmbeddingHash docs + shelved app-v20 design

l33tdawg:
Follow-up to #72, for **11.8.2**. **No consensus behaviour change** — the FinalizeBlock delegated-proof call is byte-untouched, so AppHash/LastResultsHash are identical at every height on every chain (replay-safe by construction).

## Why no fork (deep adversarial design pass, converged, 0 breaks)
Forking to tighten the v11.7.6 delegated-proof relaxation is **not worth it**:
- **R1 (node-authoritative `EmbeddingHash`) — inert on-chain.** Write-only (only presence-guards read it); everything the agent authorizes is byte-bound via `ContentHash`; only the submitting node (already inside the delegated trust boundary) can stamp a value; embeddings are per-node search index re-derived from agent-signed content.
- **R2 (dropped timestamp upper bound) — low.** Replay-marker horizon is airtight (`expiresAt = ts+skew`); the only residuals need the agent's **own** key. HTTP boundary already enforces a two-sided window.
- **No cross-chain exposure:** v11.8 sync-relays sign proof + outer tx with the **same** operator key, so `PublicKey == AgentPubKey` short-circuits the delegated gate entirely.
- A fork would spend the next rung, add a forever replay boundary, need AppHash-exclusion surgery, and **re-create the v11.7.6 idle-chain regression** at cutover.

## What this ships (zero consensus)
- **CheckTx-only future cap** in `enforceDelegatedAgentProof` (`!claim` path): rejects far-future stockpiled proofs at mempool admission. Never on the consensus path. `TestDelegatedProofMempoolFutureCapRejectsOnlyOnCheckTx` freezes that claim=true still accepts a far-future proof.
- **`EmbeddingHash` documented** as node-derived, lossy, unverified per-node search-index metadata (integrity anchor is `ContentHash`).
- **Decision record + shelved app-v20 "anchored-nonce" design + revisit triggers** (`docs/reference/concepts/delegated-agent-proof-h2-disposition.md`) — build the fork only if a trigger fires (e.g. any flow that spreads an agent key across chains).

## Verification
`golangci-lint run` = 0; `go build ./...`, `go vet ./...`, `go test ./internal/abci ./internal/tx` all green.

### DIFF
--- a/docs/reference/concepts/delegated-agent-proof-h2-disposition.md
+++ b/docs/reference/concepts/delegated-agent-proof-h2-disposition.md
@@ -1,0 +1,98 @@
+# Design record: delegated-agent-proof hardening (H-2 residual) — SHELVED for app-v20
+
+**Status: WON'T-FIX-YET (decision), with a pre-specified design shelved for a future fork.**
+This records why the v11.8 security review's H-2 residual is *not* being closed with a
+consensus fork now, what was shipped instead (11.8.2), and the exact app-v20 design to
+build if a revisit trigger fires.
+
+## Background
+
+`enforceDelegatedAgentProof` (internal/abci/agent_proof.go) enforces, in the FinalizeBlock
+consensus path, that a delegated tx (`PublicKey != AgentPubKey`) carries a valid agent
+signature over the action. v11.7.6 ("restore reliable MCP turns", 534b6fd) **relaxed** two
+rules, **ungated**:
+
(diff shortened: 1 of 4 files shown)


l33tdawg:
Closing without merge. The proposed app-v20 anchored-nonce design is not approved and app version 20 remains unallocated. The already-merged PR #72 replay-safety guard is the retained disposition.

PR #142, opened 2026-08-04
feat: finish v11.17.4 recovery, Messages, and macOS updates

l33tdawg:
## Outcome

Stacks the v11.17.4 release work on #141 without changing the stable v11.17.2 line.

- makes Messages/Inbox the canonical local and federated agent workflow (sage_message_send, sage_inbox, sage_message_reply, sage_message_status, sage_message_history); legacy pipeline tools and IDs remain readable compatibility only
- adds caller-bound atomic idempotency for federated sends and msg-* IDs for new local/federated messages
- enables signed macOS in-place update plus coordinated click-to-restart using a separately signed external recovery helper, bounded launch grace, startup proof, and launch-failure rollback
- fixes the recovery panel grid so explanatory/yellow copy spans the panel instead of being squeezed between selectors
- reports successful deprecation as Recovery Decision (local) in Activity and states clearly that preserved history was not deleted or rewritten and no chain transaction was created
- retains #141's protected Consensus scopes endpoint and app-v25 unrelated-owner prefilter, which stop the /v1/scopes 401 page failure and repeated code-72 continuity proposal loop

## Validation

- go test ./... -count=1
- changed-package rerun: REST/store/federation/MCP/web/sage-gui/update-helper
- focused race detector: federated message idempotency and canonical MCP Inbox/Messages flows
- 97 CEREBRUM, federation-route, and release-workflow JavaScript checks
- macOS packaging shell syntax and staged helper signature assertions
- disposable Docker node with two separately approved agent keys: send -> idempotent retry -> Inbox claim/read -> reply -> sender status -> node restart -> persisted completed/read-confirmed status
- inherited #141 Docker topology evidence: same-LAN direct, isolated bridge/direct routing, forced secure relay, and restart persistence

## Honest remaining boundary

Docker exercises the same encrypted direct/relay code paths and simulated network separation. A physical Mac mini roaming between real external networks remains release acceptance evidence, not a different code path or an open implementation dependency.


l33tdawg:
Docker acceptance follow-up for commit `3437664`:

- Reproduced a real canonical messaging blocker: `sage_message_send` includes `idempotency_key` in the exact signed `/v1/pipe/send` body, while the federated peer's strict proof decoder still expected the legacy body and rejected delivery with `invalid pipeline agent proof`.
- Fixed the peer decoder without carrying the sender-local idempotency token into the federated event.
- Added a regression using a token-bearing canonical request.
- Marked the four legacy MCP pipe tools explicitly deprecated in their advertised descriptions; compatibility remains for older clients and historical IDs.
- Two persisted Docker nodes on the exact patched binary completed the public workflow with no legacy MCP pipe calls:
  - `sage_message_send` A -> B returned `msg-7bfcb241-96c6-4473-ab36-a160ccf4af2b`.
  - Exact retry returned the same ID with `idempotent_replay:true`.
  - B received `msg-fed-7c37e87bd0bb61ffcf1bc2e8aa2554c8` through `sage_inbox`.
  - B replied through `sage_message_reply`; A received `ACK-11174 canonical Inbox round trip received`.
  - `sage_message_status` reported `transport_status:delivered`, receipt-v2 claim/read confirmed, and `peer_operator_durable_admission`.
  - After restarting B, `sage_message_history(folder=inbox)` retained the completed request and reply.
- `go test ./...` passes on the updated branch.

The recovery screenshot's collapsed yellow copy/two-select overflow remains covered by the responsive grid fix already in this PR.

PR #181, opened 2026-08-11
feat(cerebrum): agent connectome endpoint — neurons + weighted synapses

ihubanov:
## What

Adds **`GET /v1/dashboard/network/synapses`** — the CEREBRUM dashboard's agent *connectome*: the swarm returned as a graph of **neurons** (registered agents) and **synapses** (directed agent→agent message-bus edges, weighted by traffic).

```json
{
  "neurons":  [ { "agent_id": "…", "name": "alice", "role": "member", "domain": "…" } ],
  "synapses": [ { "from_agent": "…", "to_agent": "…", "count": 42, "last_fired": "2026-…Z" } ]
}
```

## Why

The existing `/memory/graph` visualizes *what's known* (memories + links). This is the complementary layer — *who's talking to whom* — computed from data the node already persists: the `pipeline_messages` bus and the on-chain agent registry. It's the data foundation for an "agent brain" view where each agent is a neuron and each communication channel a synapse whose weight grows with traffic.

## How

- **`internal/store` — `GetPipeSynapses`**: aggregates `pipeline_messages` by `(from_agent, to_agent)` → `COUNT(*)` and `MAX(created_at)`. Local edges only — federated cross-chain rows (`from_provider`/`to_provider` set) are excluded, so this is the local connectome. The **raw count** is returned rather than a server-computed decay, so a client can apply its own time-decay ("use it or lose it") against `last_fired`. Exposed via an **optional interface** (like `rerankerInfoProvider`), so a third-party/test store that doesn't implement it simply yields an empty connectome instead of failing.
- **`web` — `handleSynapses`**: joins those edges with `ListRegisteredAgents` for the neuron set (isolated agents included — a neuron with no connections is still a neuron). **RBAC mirrors the memory-graph edge guard exactly**: a synapse is returned only when *both* endpoints are visible to the caller, so no edge can reveal an agent the caller couldn't otherwise see; a human dashboard (no agent identity) sees all. Same `cerebrumOperatorGate` + `appV23ProjectionBroadReadGate` as `/memory/graph`.

## Determinism / safety

Pure read-aggregation. No Go consensus path touched — nothing in `internal/abci`, `internal/tx`, or `FinalizeBlock`, nothing written. It only reads two existing tables and reuses the existing dashboard RBAC.

## Tests

- `internal/store` — aggregation: count, direction (`A→B` distinct from `B→A`), federated-exclusion.
- `web` — the endpoint end to end: `{neurons, synapses}` shape, weights, arrays never null.
- `go build ./…`, `go vet`, `gofmt` clean.

## Context (optional)

This is intended as step 1 of a larger CEREBRUM "connectome" evolution — a living agent-brain view where synapses thicken with traffic (Hebbian plasticity), new agents appear as neurogenesis, and memories re-anchor to their authoring neuron so a corroborated memory reads as a *distributed engram* across cells. Happy to sketch the full arc if it's of interest — but this endpoint stands on its own as an additive read.


l33tdawg (changes requested):
Thanks for this, and apologies for the slow first response — we were mid-release on v11.18.8 and v11.18.9 and this sat longer than it should have.

**We want this.** The framing in #182 is genuinely good, the implementation is additive (+266/-0, no deletions), and it lands on data the node already persists without opening a new consensus surface. Requesting changes rather than merging, with five specific conditions below — none of them are objections to the design.

### What we checked and found sound

- **Route registration really does mirror `/memory/graph`** — `r.With(h.cerebrumOperatorGate, h.appV23ProjectionBroadReadGate)`, verified against the diff rather than the description.
- **The edge guard is correct as written.** `web/synapse_handler.go:80` returns an edge only when *both* endpoints are visible, so no synapse can reveal an agent the caller couldn't otherwise see.
- **`seeAll` for an identity-less caller is our existing convention**, not something this PR invents — `resolveAgentRBAC` at `web/handler.go:399` ("Human dashboard — no filtering"), and it sits behind `cerebrumOperatorGate`.
- **The SQL is static**, checks `rows.Err()`, and every column compared against `''` is `NOT NULL DEFAULT ''` in the schema — so there's no NULL-semantics trap silently dropping local rows.
- **Optional-interface degradation matches the existing `rerankerInfoProvider` pattern.**
- It still merges cleanly onto current `main` after two releases, and none of the five files it touches were modified by either release.

### Merge conditions

**1. Restricted-caller RBAC coverage — the one we'd most like closed.**

This is the PR's headline security property and it currently has no test. We deleted the edge guard at `web/synapse_handler.go:80-82` entirely — making every caller see every synapse — and both `./web` and `./internal/store` still passed. The existing tests cover aggregation, edge direction, federated exclusion and response shape, but none exercises a caller with a restricted view.

The guard is right today; nothing stops it being removed tomorrow. Please add a test with a restricted (non-`seeAll`) caller asserting that an edge touching a non-visible agent is filtered out — and ideally one that fails if the guard is deleted, which is the property that actually matters.

**2. A partial covering index for the aggregation.**

`GetPipeSynapses` runs `GROUP BY (from_agent, to_agent)` over the whole `pipeline_messages` table on every request, on a dashboard designed to be polled. The existing indexes don't serve it — `idx_pipe_from_agent` is `(from_agent, status)`, so the `GROUP BY` gets no covering index and the `!= ''` predicates aren't selective.

We'd rather have the index than a cache in this PR. Something like this, alongside the other `idx_pipe_*` definitions in `internal/store/sqlite.go`:

```sql
CREATE INDEX IF NOT EXISTS idx_pipe_synapse_local
  ON pipeline_messages(from_agent, to_agent, created_at)
  WHERE from_agent != '' AND to_agent != '' AND from_provider = '' AND to_provider = ''
```

The partial `WHERE` needs to match the query predicate for SQLite to use it. That covers the whole aggregation — group keys from index order, `MAX(created_at)` and the count straight from index rows. A regression test asserting the query plan uses it would be welcome where practical.

(To be explicit about the road not taken: `/memory/graph` is memoised by `graphCache` with stale-while-revalidate. We're deliberately *not* asking you to replicate that here — the index is the better fit for this shape, and adding a second cache is our architectural call to make later, not yours to guess at.)

**3. Coherent behaviour when `BadgerStore` is nil.**

`resolveAgentRBAC` returns `(nil, true)` when `h.BadgerStore == nil`, so `seeAll` is true. The neuron loop is then skipped by its own nil guard at `web/synapse_handler.go:53`, but the synapse loop still returns *every* edge — producing zero neurons alongside a full edge set referencing agents absent from the neuron list. Probably unreachable in production, but the guard you wrote for neurons should carry to synapses so the response can't be internally inconsistent.

**4. Fail loudly on a `ListRegisteredAgents` error.**

At `web/synapse_handler.go:54` the error is swallowed (`if agents, err := ...; err == nil`), so a Badger failure returns an **empty neuron list** — indistinguishable from "no agents are registered". Meanwhile a synapse read failure returns a loud 500. The dashboard would render a plausible empty brain on infrastructure failure.

This is a failure class we care about across the codebase: a capability gap must not be reportable as a legitimate zero. Please make the two paths symmetric.

**5. The weight semantics need correcting — this one is a factual fix, not a preference.**

`PipeSynapse.Count` is documented at `internal/store/pipe_synapses.go:14` as "the total number of messages **ever sent** on this directed edge". That isn't accurate, and the way it's inaccurate is non-obvious.

`pipeline_messages` rows are pruned — `internal/store/sqlite.go:6805` deletes terminal rows past a retention cutoff. But look at the predicate: `AND pipe_id NOT LIKE 'msg-%'`. **Canonical `msg-` messages are exempt from pruning; legacy pipe rows are not.** So the weight isn't one semantic with a caveat — it's two semantics sharing a number. Canonical edges accumulate indefinitely, legacy edges decay out of the table on the retention cutoff. Two edges showing the same count can mean materially different things depending on which transport produced them.

We're fine accepting bounded retained-traffic semantics for v1 — this doesn't need to become a durable counter. But the field comment and any API documentation should say **currently retained local message rows**, not "total ever", so nobody downstream builds decay math on an assumption the storage layer doesn't honour.

### Process notes

- Please don't rebase yet — land the changes above first, then rebase onto current `main` and let CI run against that. The current green checks are from `2026-08-11`, before both releases, so they're green against a merge base that's no longer `main`.
- We'd rather you make these changes than have us rewrite your branch.

### On #182

Leaving the issue open — the neuron/synapse framing is a good direction and the "free acceptance test for attribution" observation is sharp.

To be clear about scope though: accepting this PR approves **the additive read endpoint only**. It isn't an approval of the rest of the ladder — live SSE firing, persistent plasticity/pruning state, or re-anchoring memories to `submitting_agent`. That last one in particular is a change to the memory view's data model and materially more invasive than an endpoint; it needs its own discussion rather than arriving as a consequence of this one. Happy to have that conversation on the issue.

Thanks again — this is a good contribution and we'd like to get it in.

l33tdawg:
Correction on process — please **don't spend time on the above**. We'd rather not bounce this back and forth for the sake of it, so we're implementing the five items ourselves and pushing them to this branch (`maintainerCanModify` is on). The PR stays yours and your commit stays as-is; ours land on top.

Leaving the review up because the reasoning is worth having on the record, but treat it as "here's what we changed and why" rather than a request. Will follow up here when it's pushed.

ihubanov:
Sounds great — I'll hold the branch exactly as-is, all yours to land the five on top. Thanks for the unusually thorough review; the retained-vs-"total ever" distinction (#5) is a genuinely good catch — I'd missed that `msg-%` rows are pruning-exempt while legacy rows aren't. And noted on scope: this is the additive endpoint only; happy to take the memory-reanchor conversation to #182 separately when there's appetite.

l33tdawg:
Pushed `8d76ced6` onto this branch — all five items are closed. Your commit is untouched; ours sits on top.

**Each fix is mutation-verified** — reverting any one of the five makes its own test fail, so none of this is vacuous coverage.

| # | Change |
|---|---|
| 1 | Restricted-caller RBAC test. Asserts each specific leak the guard prevents (hidden `to_agent`, hidden `from_agent`, visible→hidden, hidden→hidden) and fails if the guard is deleted. |
| 2 | `idx_pipe_synapse_local`, partial + covering, carrying `created_at`. |
| 3 | Registry is now a precondition for the whole response — 503 rather than orphan edges. |
| 4 | `ListRegisteredAgents` error → 500, symmetric with the synapse read. |
| 5 | `Count` documented as retained rows. |

**One thing worth flagging, because it changed the shape of #2.** The index on its own would have been a no-op. SAGE never runs `ANALYZE` or `PRAGMA optimize`, so `sqlite_stat1` doesn't exist on a production node — and with no statistics the planner ignores the new index and instead seeks `idx_pipe_to_provider` on `to_provider = ''`, which matches nearly every local row, then sorts through a temp b-tree. Measured, not assumed:

```
no hint : SEARCH pipeline_messages USING INDEX idx_pipe_to_provider (to_provider=?) | USE TEMP B-TREE FOR GROUP BY
hinted  : SCAN pipeline_messages USING INDEX idx_pipe_synapse_local
```

So the query carries an `INDEXED BY` hint, with a fallback to the unhinted form — `migratePipeline` creates indexes best-effort and discards the error, and a slower connectome beats a broken one if the index is ever absent. The plan test deliberately runs **without** `ANALYZE`, since running it would test a database shape SAGE never actually has and would hide the fact that the plan depends on the hint.

On #5, the retention detail is sharper than "counts get pruned": `internal/store/sqlite.go` prunes terminal rows past a cutoff but exempts canonical `msg-` messages. Canonical edges accumulate; legacy edges decay out of the table. Two edges with the same weight can summarise different spans of history, which is why the field says retained rows rather than a lifetime count.

Local: `./web` and `./internal/store` full suites green, `-race` green on the changed surfaces, `go vet` and `gofmt` clean. Merging once CI is green — thanks again for the contribution, and for the framing in #182.

l33tdawg (approved):
All five requested conditions are now closed on exact head a96d642695073bdfe84c08f6e4cfff684eda3f98. Independently verified the focused tests and RBAC mutation; current-main CI is fully green, including the rerun of the app-v26 real Comet/ABCI fault gate. Approved for merge.

Issue #182, opened 2026-08-11
Proposal: evolve CEREBRUM from an MRI into a living connectome (agents = neurons, bus = synapses)


ihubanov:
# Proposal: evolve CEREBRUM from an MRI into a living connectome

The 3D brain today visualizes **what's known** — memories as nodes, links as a static connectome. With the message bus (agent↔agent over `/v1/pipe/*`), CEREBRUM now has the piece it never had: the **neurons and synapses themselves**. That's the leap from an MRI (static structure) to an **fMRI** (a brain you watch think). Every metaphor here lands on data the node already persists — no new consensus surface.

## The mapping

| Neuroscience | SAGE primitive |
|---|---|
| **Neuron** | a registered agent (`network_agents`) |
| **Synapse** | a directed `/v1/pipe` channel between two agents |
| **Synaptic weight** | message count on that edge, time-decayed |
| **Hebbian plasticity** ("fire together, wire together") | weight ↑ per message, decays when silent |
| **Synaptic transmission** | pipe lifecycle `pending→claimed→completed` |
| **Neurogenesis** | `AgentRegister` (an agent-builder spawns a neuron) |
| **Pruning / apoptosis** | agent removed, or `last_seen` goes cold |
| **Distributed engram** | a memory corroborated by N agents = held across N neurons |

## The two-layer brain (keeps the memory viz, doesn't replace it)

Memories and agents aren't competing views — they're two layers of one organ. Neurons + synapses are the *wiring*; memories are the *engrams encoded on it*. So the memory brain doesn't get thrown away — it gets **re-homed**: anchor each memory to the agent that authored it (`submitting_agent`), and a memory stops floating and becomes an engram orbiting its neuron. Zoom out → the living connectome; zoom into a neuron → its engrams bloom.

And the payoff **completes the original CLS analogy** rather than abandoning it: a memory corroborated by many agents becomes a **distributed engram across many neurons** — which is *literally* how the brain consolidates long-term memory. Consolidation, which the current brain shows as radial depth, now also shows as distribution across cells. Two views of the same truth.

## The PR ladder (small, additive, dependency-ordered)

1. **`/network/synapses` endpoint** — neurons + weighted synapses, RBAC'd like `/memory/graph`. *(PR #181, open now — the working foundation)*
2. **"Agent brain" view** — render agents as neurons, synapses as edges, in the existing brain hull; thickness/brightness = weight.
3. **Live firing (fMRI)** — synapses pulse as messages transit, over SSE.
[...]

PR #188, opened 2026-08-12
feat(cerebrum): connectome view — render the agent message-bus in the brain hull

ihubanov:
## What

Adds an additive **connectome** view mode to the CEREBRUM 3D brain. A HUD toggle button (◉ connectome) swaps the existing brain between the memory-graph view (unchanged) and a new agent-connectome view backed by the merged `/v1/dashboard/network/synapses` endpoint (#181): each registered agent renders as a neuron, each directed agent→agent channel as a synapse whose thickness, particle count and pulse speed scale with traffic (Hebbian weight). Hub agents sink toward the brain core, peripheral agents sit on the cortex, neurons colour by domain. Clicking a neuron lights its synaptic neighbourhood.

The 3D result is best judged live — opening this for your eyes before the further rungs land.

## Why

`/v1/dashboard/network/synapses` (#181) exposes the agent message-bus but nothing renders it; the operator can query the topology but can't see it. The memory graph already carries a deterministic brain-hull renderer with placement, focus-fade and tip pipelines. Rather than stand up a second visualization, the connectome is projected onto that same render contract so one code path serves both views.

Step 2 of the connectome arc in #182. The further rungs — live SSE firing, persistent plasticity, memory re-anchoring — remain separate and discussion-first, and are **not** in this PR.

## How

- `web/static/js/connectome-map.js` — new pure mapper, dependency-free (no DOM, no `fetch`, no THREE), mirroring the `mri-layout.js` pattern. Projects `{neurons, synapses}` onto the internal `{nodes, links}` contract the memory graph renders. Neuron degree `_deg` (share of the busiest neuron's total in+out traffic, 0..1) drives size and radial depth; synapse weight `_w` (share of the busiest edge's count) drives Hebbian thickness/pulse. Neurons carry neutral values for the memory fields the shared tip/render paths read.
- Ghost-edge defence: any synapse whose endpoints are not both in the neuron set is dropped, and a dropped edge does not leak into neuron traffic weight. The server already both-endpoints-gates synapses by RBAC; this is defence-in-depth so the renderer never mints a node with no deterministic placement.
- `web/static/js/mri-brain.js` — all new behaviour gated on a `mode` flag (`'memory'` default). `placeNeurons()` reuses the same deterministic ellipsoid: domain → azimuthal lobe, and **degree → radial depth** via the same `mriDepthForAge`/`mriVerticalPosition` scale age uses for memories, so both views share one spatial language. Link width/particles/speed become `_w`-aware accessors; the synapse link type is styled dynamically while memory link formulas are untouched. `focusNeuron()` lights the clicked cell plus every agent it exchanges with by reusing the memory view's `focusSet` fade — no network fetch, purely the already-loaded graph. Stat panel relabels to neurons/synapses/hubs; lobes become a colour legend (the connectome endpoint is not domain-drillable); a bottom-left caption explains the mapping.
- `resolveConnectomeLinks()` — the renderer pins node positions and nulls the d3 forces, so ForceGraph3D never runs the lazy link-id resolution the `'link'` force performs. For the connectome the synapses **are** the payload, so each edge's endpoints are resolved to their node objects before `graphData()`; without this every synapse references an unresolved id and renders nothing. Scoped to connectome mode. (Note for the maintainer: the memory view's typed links appear to share this property under the pinned layout — flagging as a possible separate follow-up, out of scope here.)

## Determinism / safety

Frontend-only. No Go, no `internal/`, no consensus path touched — this is static JS reading one already-merged read endpoint. Purely additive and reversible: the memory view is byte-for-byte unchanged whenever `mode` stays `'memory'`, and the whole feature is gated behind `opts.allowConnectome` (default on) and the `mode` flag. A transport/parse failure on the synapses fetch is surfaced (via `loadSynapses`), not rendered as an empty brain, matching `loadGraph`.

## Tests

- `scripts/connectome-mapping.test.mjs` — 7 cases pinning the mapper invariants the renderer relies on: degree + weight normalization against the busiest neuron/edge, field fallbacks, direction preserved (A→B distinct from B→A), ghost-edge rejection (dropped edge leaves zero traffic), and a safe empty connectome for `null`/`undefined`/`{}`/malformed payloads. Registered in the npm `test` list.
- `npm test` green — `check:js` plus the full node suite, 268 tests pass.

## Verification

Beyond the unit + `npm test` suites, I rendered the view in headless Chromium (SwiftShader WebGL) against a synthetic swarm to confirm the 3D result — neurons place by domain with hubs pulled to the core, and synapses render as weighted channels fanning out from the hubs. That harness is what surfaced the `resolveConnectomeLinks` fix above (synapses were initially invisible under the pinned layout).

## Context

Depends on #181 (already merged). Part of the connectome arc in #182; subsequent rungs are tracked separately there. This is a draft — the logic is unit-tested and the render sanity-checked headless, but the polished 3D look is still best eyeballed live on real data.


ihubanov:
Thanks for landing this — and for the `createGraphLoadCoordinator` fix especially. You're right: my in-flight guards only checked disposed/Graph, not that the mode still matched, so a slow load resolving after a toggle would paint stale-mode data into the other view. Stamping each load with its request mode and dropping stale resolutions is the correct fence, and the race + wiring tests are a better safety net than I left. (The caption move to the right was the right call too — it was clipping the HUD.)

Both rungs are up now. The remaining arc — live SSE firing, plasticity, the memory→submitting_agent re-anchor — stays parked on #182 for whenever there's appetite; happy to take any of them behind your steer, or leave them for you.

PR #196, opened 2026-08-13
fix(cerebrum): key the local connectome on chain ids, not provider labels

l33tdawg:
**Draft — not for merge without codex's review.** Codex holds merge authority and is independently verifying this finding.

## The defect

`GetPipeSynapses` restricted the connectome to rows with empty `from_provider`/`to_provider`, on the stated premise that federated cross-chain rows are the ones carrying a provider. **That premise is false in both directions.**

`from_provider`/`to_provider` hold the **agent's provider label**, not a federation marker. `SAGE_PROVIDER` flows from the generated MCP launcher into agent registration (`internal/mcp/tools.go:3114`) and is stamped onto every send (`api/rest/messages_handler.go:94-96,:114`), so ordinary local traffic arrives tagged `claude-code` or `codex`.

Meanwhile the federated send path does the opposite of what the comment assumed — it **clears** `ToProvider` and sets `DestinationChainID`:

```go
// api/rest/pipe_handler.go:548-549
req.ToProvider = ""
req.DestinationChainID = remoteTarget.ChainID
```

So the old predicate **hid local traffic** and **admitted federated traffic**.

## Measured on a live node, before this change

| | |
|---|---|
| local agent-to-agent rows | 363 |
| excluded by the old predicate | **359** |
| edges the endpoint returned | **3** |
| …of which were actually cross-chain, counted as local | **1** |
| genuine local edges hidden | 26 |

Provider distribution, all with no chain ids so unambiguously local: `codex→(empty)` 206, `claude-code→(empty)` 133, `audit→(empty)` 20, `(empty)→(empty)` **4**.

## The fix

Locality keys on `source_chain_id`/`destination_chain_id`, which is what actually denotes a cross-chain row. Both are `NOT NULL DEFAULT ''`, so there is no NULL-semantics trap (verified: 0 NULLs across 404 live rows).

**The index `DROP` is deliberate and load-bearing.** `CREATE INDEX IF NOT EXISTS` is a no-op against an index that already exists under the same name with the old partial `WHERE`. Without the drop, an upgraded v11.18.9 node keeps a stale partial index that no longer matches the query, SQLite declines to use it, and the `INDEXED BY` hint quietly degrades to a full scan on a table that only grows. Same drop-then-recreate precedent as `idx_pipe_transport_proof_once`.

## The old test agreed with the bug

The existing fixture built its "federated" row as `fed.ToProvider = "other-chain"` — which production never writes. Test and code shared one wrong assumption and so confirmed each other. It now models federation the way the federated send path actually writes it.

## Coverage, each mutation-verified

| test | mutation that breaks it |
|---|---|
| `…IncludesProviderBearingLocalEdges` | revert to the provider predicate |
| `…StillExcludesCrossChainEdges` | revert to the provider predicate (**proves the inclusion half**) |
| `…UsesCoveringIndex` | revert predicate, or drop the `INDEXED BY` hint |
| `…IndexIsRedefinedOnUpgrade` | remove the `DROP INDEX` |

Reverting the predicate fails **both** the inclusion and exclusion assertions — the leak is demonstrated, not argued.

## Gates

`./internal/store` full suite green (102s) · `./web` synapse tests green · `-race -count=2` green · `go vet`, `gofmt`, `go build ./...` clean.

## Scope

Deliberately **not** included: the app-v23 `seeAll` question, the existing global-stream leak, and event-wiring coverage — all reported separately and all codex's call. This PR changes one predicate and its index.

### DIFF
--- a/internal/store/pipe_synapses.go
+++ b/internal/store/pipe_synapses.go
@@ -38,7 +38,7 @@ type PipeSynapse struct {
 // The %s slot carries an INDEXED BY hint. It is needed rather than decorative:
 // SAGE never runs ANALYZE or PRAGMA optimize, so sqlite_stat1 does not exist on
 // a production node, and without statistics the planner prefers a seek on
-// idx_pipe_to_provider (to_provider=”) — which matches nearly every local row —
+// idx_pipe_to_provider on an empty to_provider — which matches nearly every local row —
 // and then sorts through a temp b-tree. Measured on an unanalysed database the
 // hint is the difference between that and a covering index-order scan.
 const pipeSynapseAggregation = `
@@ -64,7 +64,7 @@ const pipeSynapseIndexHint = "INDEXED BY idx_pipe_synapse_local"
 // almost every real local edge: on a live node it returned 3 of 30 edges and
 // hid 359 of 363 local rows. source_chain_id/destination_chain_id are the
 // fields that actually denote a cross-chain row, and both are NOT NULL
-// DEFAULT ”, so the comparison has no NULL-semantics trap.
+// DEFAULT with an empty string, so the comparison has no NULL-semantics trap.
 func (s *SQLiteStore) GetPipeSynapses(ctx context.Context) ([]PipeSynapse, error) {
 	// migratePipeline creates indexes best-effort and discards the error, so the
 	// hint must not become a hard dependency: an INDEXED BY naming an absent
--- a/internal/store/pipe_synapses.go
+++ b/internal/store/pipe_synapses.go
@@ -45,7 +45,7 @@ const pipeSynapseAggregation = `
 	SELECT from_agent, to_agent, COUNT(*), MAX(created_at)
 	FROM pipeline_messages %s
 	WHERE from_agent != '' AND to_agent != ''
-	  AND from_provider = '' AND to_provider = ''
+	  AND source_chain_id = '' AND destination_chain_id = ''
 	GROUP BY from_agent, to_agent`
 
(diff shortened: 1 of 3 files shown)


ihubanov:
Independently verified the mechanism against origin/main, and the finding holds — this was my premise error in #181, wrong in both directions:

- `api/rest/pipe_handler.go:548-549` — the federated branch `req.ToProvider = ""` then sets `DestinationChainID`, so cross-chain rows carry an **empty** provider. My `to_provider = ''` predicate therefore *admitted* federated rows.
- `api/rest/messages_handler.go:114` — local sends stamp `FromProvider: fromProvider`, so ordinary local traffic is tagged `claude-code`/`codex`. My `from_provider = ''` predicate therefore *excluded* almost every real local edge.
- `internal/store/sqlite.go:6133/6135` — `source_chain_id`/`destination_chain_id` are `NOT NULL DEFAULT ''`, so keying locality on the chain ids has no NULL trap. That's the right discriminator.

I can't reproduce your live 3-of-30 / 359-of-363 counts from here, but the code mechanism that produces them is exactly as described. I should have traced the federated send path before asserting the comment's premise — I inverted what the provider fields mean. Thanks for measuring it on a real node and catching it.

The `DROP INDEX` before recreate is right too: `idx_pipe_synapse_local`'s partial `WHERE` was written for the old predicate, and `CREATE INDEX IF NOT EXISTS` would leave the stale partial in place on an upgraded node, silently degrading the `INDEXED BY` hint to a full scan.

No change needed on the viz side (#188) — it renders whatever the endpoint returns, so it fills in automatically once this lands. Happy to help however's useful.

PR #197, opened 2026-08-13
fix(cerebrum): open the connectome with a dimmer default skull

ihubanov:
## What

Connectome mode now opens with a dimmer default skull (0.03) than the memory graph (0.08), so the synapse wiring reads immediately instead of being drowned by the bright anatomical wireframe.

## Why

The connectome's value is the wiring — agents as neurons, message-bus channels as weighted synapses. At the memory view's default skull opacity the dense hull mesh visually dominates the thinner synapses, so on first open the operator mostly sees wireframe and has to reach for the SKULL slider before the graph's point is legible. The two views want different defaults: the memory graph benefits from a prominent anatomical hull (dense nodes, age-depth); the connectome from a recessed one.

## How

- `MODE_HULL = { memory: 8, connectome: 3 }` (slider units; value/100 = opacity). `curOpacity` initializes from the active mode, the slider reflects it from mount, and `setMode()` applies the target view's default (and updates the slider) on toggle.
- Still fully user-adjustable via the SKULL slider from either default.

## Determinism / safety

Frontend-only, connectome-scoped. No Go/consensus path touched. The memory view is byte-for-byte unchanged (still 0.08).

## Tests

`npm test` green (272). Verified headless (Chromium/SwiftShader): the connectome opens with synapses legible against a recessed hull, hubs at the core, neurons by domain; memory view unchanged.

## Context

Follows #188 (connectome view). Aesthetic default — happy to tune the exact value to your eye; the per-view mechanism is the point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "check:js": "node scripts/check-static-js.mjs",
-    "test": "npm run check:js && node --test scripts/access-controls-ui.test.mjs scripts/appv23-linked-message-ui.test.mjs scripts/appv23-policy.test.mjs scripts/bulk-sequence.test.mjs scripts/cerebrum-shell.test.mjs scripts/chrome-extension-contract.test.mjs scripts/connectome-mapping.test.mjs scripts/domain-inventory.test.mjs scripts/federation-acceptance-contract.test.mjs scripts/federation-flow.test.mjs scripts/federation-peer-capabilities.test.mjs scripts/federation-route-state.test.mjs scripts/filter-codeql-sarif.test.mjs scripts/governance-retry.test.mjs scripts/release-recovery-workflow.test.mjs scripts/release-workflow.test.mjs scripts/restart-proof.test.mjs scripts/task-reorder.test.mjs scripts/update-banner.test.mjs scripts/version-alignment.test.mjs"
+    "test": "npm run check:js && node --test scripts/access-controls-ui.test.mjs scripts/appv23-linked-message-ui.test.mjs scripts/appv23-policy.test.mjs scripts/bulk-sequence.test.mjs scripts/cerebrum-shell.test.mjs scripts/chrome-extension-contract.test.mjs scripts/connectome-mapping.test.mjs scripts/mode-hull.test.mjs scripts/domain-inventory.test.mjs scripts/federation-acceptance-contract.test.mjs scripts/federation-flow.test.mjs scripts/federation-peer-capabilities.test.mjs scripts/federation-route-state.test.mjs scripts/filter-codeql-sarif.test.mjs scripts/governance-retry.test.mjs scripts/release-recovery-workflow.test.mjs scripts/release-workflow.test.mjs scripts/restart-proof.test.mjs scripts/task-reorder.test.mjs scripts/update-banner.test.mjs scripts/version-alignment.test.mjs"
   },
   "repository": {
     "type": "git",
--- a/scripts/mode-hull.test.mjs
+++ b/scripts/mode-hull.test.mjs
@@ -1,0 +1,111 @@
+import assert from 'node:assert/strict';
+import { readFile } from 'node:fs/promises';
+import test from 'node:test';
+
+import { createModeHull } from '../web/static/js/mode-hull.js';
+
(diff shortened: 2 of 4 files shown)


l33tdawg (changes requested):
Thanks — the per-mode visual defaults are useful, but this head is not release-ready yet. Two blockers: (1) `setMode()` reapplies 0.03/0.08 on every toggle, so an operator’s manual SKULL slider choice is discarded on a round trip; please retain per-mode session values, using 0.03/0.08 only as each mode’s initial default. (2) no regression was added: deleting the entire change leaves the existing suite green. Add a non-vacuous behavior/source test covering initial memory/connectome values, slider reflection, manual adjustment persistence across mode toggles, and independent per-mode state. Then update onto current main (`ab4523b2...`) and rerun exact-head gates.

ihubanov:
Both addressed; updated onto `ab4523b2` and pushed.

**(1) Manual choices now persist per view.** The 0.03/0.08 values are initial defaults only. Per-view opacity lives in a small pure module (`web/static/js/mode-hull.js`): `setMode()` recalls the active view's remembered value (default or the operator's last manual choice) and reflects it on the slider, and the `.b-op` `oninput` handler records the value for the active view. A round trip preserves each view's setting, and the two are independent.

**(2) Real regression added** — `scripts/mode-hull.test.mjs` (registered in `npm test`): behavioural coverage of the pure state (initial defaults, persistence across a mode round trip, per-view independence, unknown-mode fallback, defaults()-copy) plus source-wiring assertions against `mri-brain.js` (createModeHull seeding, `valueFor` recall, slider reflection, `record` on manual input, curOpacity seeded from state). Reverting either the module logic or the wiring now fails a test — deleting the change no longer leaves the suite green.

Exact-head `npm test` green (281). Also re-verified the render headless: the connectome still opens dim, and a slider adjustment survives a memory↔connectome round trip. Tune the 0.03 to your eye whenever.

ihubanov:
Ready for another look. The one red check — `compile-and-contract (windows-2025 … nsis)` — fails in the *Windows package lifecycle smoke* step (a runtime install test), which a frontend-only JS change can't affect; the same job passed on #196 in parallel, so it reads as a flaky cross-platform gate rather than anything from this diff. The gates that exercise this change — Frontend Static Checks, Lint, Test — are all green on the exact head (`c616403a`). Happy to rebase for a re-run if you'd prefer a clean board.

l33tdawg (changes requested):
Amended UX behavior is correct, but one regression blocker remains at exact head c616403a: deleting `setHullOpacity(nextOpacity)` from `setMode()` leaves every new test green, while the slider changes and the visible hull incorrectly retains the prior mode opacity. The source assertions are unscoped and only pin `valueFor(mode)` plus slider assignment. Please add a scoped `setMode` wiring assertion or behavior test that binds the same recalled `nextOpacity` through both slider reflection and `setHullOpacity(nextOpacity)`; mutation deleting or substituting that application must fail. Exact-head Windows native retry is also still pending.

ihubanov:
Fixed at `239de3c8`. Added a **scoped** `setMode` wiring assertion (`scripts/mode-hull.test.mjs`): it extracts the `setMode` function body by brace-matching and asserts the recalled `nextOpacity` is bound through *both* `opEl.value = sliderUnits(nextOpacity)` **and** `setHullOpacity(nextOpacity)` — same variable, so the slider and the visible hull can't drift on a toggle.

Mutation-verified it does what you asked:
- baseline: 10/10 green
- delete `setHullOpacity(nextOpacity)` from `setMode` → the new test fails (9 pass / 1 fail)
- restore → 10/10 green

Substituting a different value (e.g. `setHullOpacity(curOpacity)`) fails the same assertion, since it's bound to `nextOpacity` specifically. Full `npm test` green (282). Re-pushed onto current main so the Windows native gate re-runs on this head.

l33tdawg (approved):
Independent exact-head re-review complete at 239de3c8e117665a2a1ea87ed7b2f5abde7eaded. Prior mutation-vacuity blocker is closed: deleting or hardcoding the visible setHullOpacity(nextOpacity) application fails the scoped regression. Focused 10/10 and full JS/static 282/282 passed independently; all hosted exact-head CodeQL, CI, native, app-v20/app-v26, Build, and Byzantine gates are green.

PR #203, opened 2026-08-14
test(tx): cover the CometBFT log scrubs directly from internal/tx

l33tdawg:
**Draft — merge authority codex's.** Base `a2763325`. **Strictly test-only**: `git diff --name-only origin/main HEAD` returns one file, and it is a `_test.go`.

First of three splits from the closed PR #200. Ordered deliberately: this one carries **zero production risk** and cannot interact with the v11.18.11 release path.

## The hole, reproduced on *current* main

`BroadcastCometCommit` and `BroadcastCometSync` scrub the CheckTx and TxResult logs before returning them, so a compromised or hostile CometBFT endpoint cannot get attacker-controlled text into operator-visible output. Those call sites were covered **only from `web/`**.

Verified today against `a2763325`, not carried over from the earlier run:

```
neuter all 3 ScrubBroadcastText call sites in comet_commit.go
  → ./internal/tx WITHOUT this test:  ok      (the hole)
  → ./internal/tx WITH this test:     FAIL    (closed)
```

A guarantee with no direct coverage in its own package is one edit away from silently disappearing — and the package that owns it would not have noticed.

## What the tests do

Drive a real broadcast against an `httptest` server returning a CheckTx/TxResult log that contains **the actual encoded transaction hex for that submission**, then assert the surfaced text does not contain it and does carry the redaction marker.

The witness being *real* matters: a fixture whose log contains nothing scrubbable makes the assertion vacuous, and that exact mistake was made on this code before.

## Mutations, all detected

| mutation | |
|---|---|
| neuter the CheckTx log scrub | FAIL |
| neuter the TxResult log scrub | FAIL |
| neuter the sync-path scrub | FAIL |
| **keep the scrub call but unbind the encoded tx** | FAIL |

That last one is the subtle case — the call still looks right, and only a test that ties the witness to the *actual* submitted bytes can see it.

## Gates

`./internal/tx` full suite green · `-race -count=2` green · `go vet`, `gofmt` clean · working tree clean.

Two more splits follow: `pipeline_send` privacy (small production change), then the SSE event-wiring guard (touches `web/sse.go`, sequenced last to stay clear of the release).

### DIFF
--- a/internal/tx/comet_commit_log_scrub_test.go
+++ b/internal/tx/comet_commit_log_scrub_test.go
@@ -1,0 +1,240 @@
+package tx
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync/atomic"
+	"testing"
+)
+
+// The CheckTx and TxResult logs that BroadcastCometCommit and BroadcastCometSync
+// hand back are remote-derived text: a compromised CometBFT endpoint, or a
+// reverse proxy in front of an honest one, chooses every byte of them. They are
+// scrubbed at the call site so a hostile node cannot echo the signed transaction
+// it was just handed back into operator-visible output.
+//
+// Until these tests existed the scrub on those three fields was pinned only from
+// the web package. Deleting all three calls from comet_commit.go left the whole
+// of ./internal/tx green, so the package that owns the scrub could not detect
+// losing it. Everything below drives the real broadcasters against an httptest
+// CometBFT and asserts on what the caller actually receives.
+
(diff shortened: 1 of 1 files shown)


l33tdawg:
Independent exact-head CODE GO recorded for `687f71100e486b6e001225d2f9a00d4db2cb9343` over base `f0bb00b0270329527ae8cfd146d51e823ad537ee`. The PR remains strictly test-only; `internal/tx/comet_commit_log_scrub_test.go` is blob `c110c8da1be665463d8abca58e48aa624d8281ea`, unchanged from the reviewed version. Focused scrub regressions and full `./internal/tx` pass, and neutering the three owned `ScrubBroadcastText` sites makes the regression fail. Merge remains gated on every exact-head hosted check, including downstream Build and Byzantine.

PR #204, opened 2026-08-14
feat(cerebrum): connectome plasticity — idle synapses thin and fade (#182 step 4)

ihubanov:
## What

Adds **resting-weight plasticity** to the connectome (step 4 of #182): a synapse's *retained* strength now decays toward a floor the longer it sits idle and climbs back as it fires, keyed on `last_fired`. An idle connectome visibly settles — thinner, dimmer channels — while active circuits stay bright. "Use it or lose it."

## Why

Today every synapse renders at its raw Hebbian weight, so an edge that carried heavy traffic weeks ago reads exactly as strong as one busy this minute. The graph shows *who has ever talked to whom*, not *which circuits are alive now*. Plasticity is the missing half of the Hebbian metaphor and the thing that makes the brain look **alive at rest**.

This composes with — and is deliberately distinct from — the live firing pulse (#201): the pulse is a ~2.6s flash on a message in flight; plasticity is the slow drift of the resting weight over minutes/hours. Both ride the same synapse accessors.

## How

- `web/static/js/connectome-map.js` — new pure `synapsePlasticity(lastFiredISO, now, {halfLifeMs, floor})`: a 0..1 multiplier that halves from 1 toward `floor` every half-life (defaults 30 min / 0.15), clamped so an idle edge dims but never vanishes and clock skew never exceeds 1. The server already returns raw `count` + `last_fired` and explicitly defers time-decay to the client (`pipe_synapses.go`: *"Time-decay is left to the client"*) — this is the client honouring that contract.
- `web/static/js/mri-brain.js` — synapse width, particle count/speed, and colour alpha render `restingWeight(l) = _w * plasticity(last_fired)` instead of raw `_w`. A self-gated 30s refresh re-reads the accessors so an idle connectome keeps settling with no new events (cleared on dispose). Tunable via `opts.plasticity`.

## Determinism / safety

Frontend-only, connectome-scoped. No Go/consensus. The memory view is byte-for-byte unchanged; the pulse path (#201) is untouched — plasticity only replaces the resting `_w` term the accessors read.

## Tests

`scripts/connectome-plasticity.test.mjs`: decay-curve coverage (half-life math, monotonic + bounded `[floor,1]`, idle→floor asymptote, missing/future `last_fired`, configurable knobs) plus **scoped** mri-brain wiring assertions binding `restingWeight` through the width and colour accessors. Mutation-verified: reverting the width term to raw `_w` fails the suite. Full `npm test` green (308).

## Verification

Rendered headless (Chromium/SwiftShader) against a synthetic swarm spanning fresh→ancient `last_fired`: fresh channels read thick and bright, idle ones (hours/days) thin and dim toward the floor — a clear at-rest gradient, with the memory view unaffected.

## Context

Step 4 of the connectome arc in #182 (after #181 endpoint, #188 view, #201 live firing). Half-life/floor are tunable — happy to adjust to your eye.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "check:js": "node scripts/check-static-js.mjs",
-    "test": "npm run check:js && node --test scripts/access-controls-ui.test.mjs scripts/appv23-linked-message-ui.test.mjs scripts/appv23-policy.test.mjs scripts/bulk-sequence.test.mjs scripts/cerebrum-shell.test.mjs scripts/chrome-extension-contract.test.mjs scripts/connectome-mapping.test.mjs scripts/mode-hull.test.mjs scripts/domain-inventory.test.mjs scripts/federation-acceptance-contract.test.mjs scripts/federation-flow.test.mjs scripts/federation-peer-capabilities.test.mjs scripts/federation-route-state.test.mjs scripts/filter-codeql-sarif.test.mjs scripts/governance-retry.test.mjs scripts/release-recovery-workflow.test.mjs scripts/release-workflow.test.mjs scripts/restart-proof.test.mjs scripts/task-reorder.test.mjs scripts/update-banner.test.mjs scripts/version-alignment.test.mjs"
+    "test": "npm run check:js && node --test scripts/access-controls-ui.test.mjs scripts/appv23-linked-message-ui.test.mjs scripts/appv23-policy.test.mjs scripts/bulk-sequence.test.mjs scripts/cerebrum-shell.test.mjs scripts/chrome-extension-contract.test.mjs scripts/connectome-mapping.test.mjs scripts/connectome-plasticity.test.mjs scripts/mode-hull.test.mjs scripts/domain-inventory.test.mjs scripts/federation-acceptance-contract.test.mjs scripts/federation-flow.test.mjs scripts/federation-peer-capabilities.test.mjs scripts/federation-route-state.test.mjs scripts/filter-codeql-sarif.test.mjs scripts/governance-retry.test.mjs scripts/release-recovery-workflow.test.mjs scripts/release-workflow.test.mjs scripts/restart-proof.test.mjs scripts/task-reorder.test.mjs scripts/update-banner.test.mjs scripts/version-alignment.test.mjs"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


l33tdawg (changes requested):
CODE NO-GO at exact head 53ef9d90f5b1db3abfb640924c5f2f857bbffe4b. The current suite does not mutation-pin three advertised production behaviors: (1) linkParticlesFor can be reverted from restingWeight(l) to raw _w, (2) linkParticleSpeedFor can be reverted from restingWeight(l) to raw _w, and (3) the 30s refresh can drop .linkDirectionalParticles(linkParticlesFor); scripts/connectome-plasticity.test.mjs still passes 12/12 in each combined mutation. Add scoped assertions/behavioral tests that both particle accessors use restingWeight with no raw _w fallback and that the timer reapplies linkDirectionalParticles. Full npm test is otherwise 308/308 and hosted gates are green on this rejected head.

ihubanov:
Fixed at `95cfe5c5`. Added the three scoped assertions and mutation-verified each against the exact mutations you named:

- **`linkParticlesFor`** must render `restingWeight(l)*5` with no `(l._w||0)*5` fallback → reverting it to raw _w: **fail** (14/1).
- **`linkParticleSpeedFor`** must render `restingWeight(l)*0.02` with no raw fallback → reverting: **fail** (14/1).
- **`startPlasticityDecay`** must contain `.linkDirectionalParticles(linkParticlesFor)` → dropping it from the refresh: **fail** (14/1).

Restored → 15/15. Full `npm test` green (311). You were right that width+colour alone left the two particle accessors and the timer's particle re-application unpinned — a reader could have silently reverted them and kept an idle synapse firing particles at full historical rate. Thanks for running the combined mutations.

l33tdawg (approved):
CODE GO at exact head 95cfe5c5e0c89220e70adb81e25c9f6b15850c0a over exact base a70e1bb9249d2f03cfe5032dbb84dcf860e68dd4. Delta from rejected head is test-only. Independent mutations reverting particle count, reverting particle speed, or removing timer particle refresh each fail the new contract. Focused plasticity+mapping 37/37 and x10 pass. Approval is exact-head code review only; merge remains blocked until every replacement-head hosted CI, CodeQL, native, consensus, Build, and Byzantine gate is green.

PR #205, opened 2026-08-14
feat(cerebrum): connectome neurogenesis + pruning (#182 step 5)

ihubanov:
## What

Adds the final rung of the brain metaphor (#182 step 5): **neurogenesis** (new agents grow in) and **pruning** (dormant agents grey out).

- **Neurogenesis** — an agent whose id was absent from the previous snapshot is stamped `_bornAt` and briefly swells + flares toward a bright accent as it grows in, then settles, so a newly-registered agent reads as a cell being born rather than blinking into place.
- **Pruning / apoptosis** — a neuron greys toward the existing challenged/deprecated grey in proportion to how long its circuit has been silent, so live agents stand out and a cold swarm visibly recedes.

## Why

The view shows every registered agent at equal presence whether it just joined or has gone silent for days. Neurogenesis and pruning are the two remaining halves of the "agent = neuron" metaphor and the thing that makes the swarm's *lifecycle* legible — who's new, who's alive, who's cold.

## How

Both are derived **client-side from data already in the payload — no endpoint change**.
- `web/static/js/connectome-map.js`: `mapConnectome` now carries per-neuron `_activity` (the most recent `last_fired` across its incident synapses). New pure primitives: `neuronDormancy(activityISO, now, {liveMs, coldMs})` → a 0..1 grey factor ramping from lively (fired within `liveMs`, default 1h) to fully dormant (idle past `coldMs`, default 12h; a neuron that never fired is dormant); and `createNeuronBirthTracker()` — first snapshot seeds the baseline, later snapshots return only newly-appeared ids (mirrors the existing `createConnectomeActivityTracker`).
- `web/static/js/mri-brain.js`: neuron colour greys by dormancy and flares on birth, `nodeVal` adds a transient birth swell, and a self-gated 30s refresh lets dormancy drift with no new events (cleared on dispose). Tunable via `opts.neuronDormancy`.

## Determinism / safety

Frontend-only, connectome-scoped. No Go/consensus. The memory view is byte-for-byte unchanged.

## Tests

`scripts/connectome-neurogenesis.test.mjs`: dormancy ramp (bounds, linearity, never-fired→dormant, configurable thresholds), birth-tracker diff (baseline seed, new-only, re-registration, reset), `_activity` mapping, plus **scoped** mri-brain wiring assertions binding the grey/birth render. Mutation-verified: stripping the dormancy grey blend fails the suite. Full `npm test` green (310).

## Verification

Rendered headless (Chromium/SwiftShader): the connectome renders cleanly with agents active within the last hour at full domain colour and agents idle for days greyed toward the deprecated grey; the effect scales with how cold the swarm is. Neurogenesis is a ~3s transient flare on a newly-appearing id, covered by the birth-tracker + wiring tests.

## Context

Completes the connectome arc in #182 (#181 endpoint, #188 view, #201 live firing, plus #204 plasticity). Independent of #204 (that PR decays edge weight; this greys nodes) though both touch the same files — happy to rebase whichever lands second. Thresholds tunable to taste.


l33tdawg (changes requested):
CODE NO-GO at exact head 3bbe1fd2f9e31f4d95fdcd2c0b279597534e5e29. Blocking findings: (1) successful agent registration emits the existing agent SSE event, but the connectome view subscribes only to remember/forget/connectome; the new 30s interval only recolors already-loaded nodes, so a newly registered silent agent cannot appear/grow until unrelated activity or reload. Add a self-gated authorized refetch on committed agent registration with cleanup/in-flight/error coverage. (2) Current tests survive deleting markConnectomeBirths(d) and replacing the actual dormancy RGB/alpha blend with no-ops; add behavioral or tightly scoped production wiring tests that fail both mutations. (3) mapConnectome compares variable-precision RFC3339Nano strings lexically, so e.g. ...00.1Z incorrectly sorts after the later ...00.12Z; parse/normalize timestamps and add a fractional-precision regression. Focused 14/14, check:js, full npm 310/310, and all hosted checks pass on this rejected head.

ihubanov:
Fixed all three at `04c86b9e` (rebased on current main).

**(1) Neurogenesis now fires live.** Added a `sse.on('agent', …)` subscription that, connectome-gated and debounced, refetches through the existing `load()` path — so a committed registration re-runs `markConnectomeBirths` and the new agent grows in. `load()` owns in-flight coalescing + the unavailable/retry path, and the debounce timer is cleaned up with the other subs, so a dropped/failed registration refetch never leaves the view mid-birth.

**(2) Both mutations now fail.** Extracted the neuron colour into a pure `neuronTint(domainRGB, deg, dorm, birth)` unit-tested by numbers (dorm=1 → exactly [108,120,145] at α0.30; birth=1 → flares to the accent at α0.9) — a no-op blend fails the numbers, not just a grep. And I now assert `markConnectomeBirths` is invoked in **both** load paths, so deleting either call site fails. (Also pinned the two timer re-applications — dormancy→nodeColor, birth→nodeVal+nodeColor — which I'd left unpinned.)

**(3) Timestamp bug fixed.** `_activity` now compares via a parsed-epoch `isLater()` instead of lexically, with a regression proving `…00.12Z` (120ms) beats `…00.1Z` (100ms) rather than losing to it lexically.

Mutation-verified each: deleting the birth call, no-op'ing the dormancy blend, dropping the dormancy recolor, removing the `agent` subscription, and reverting `isLater` to a lexical compare each fail the suite; restored → 24/24. Full `npm test` 320/320, render re-verified (no regression from the tint refactor). Thanks — the live-registration gap especially was a real hole.

l33tdawg:
Additional exact blocker in the same owned file: the shipped #201 diffConnectomeActivity path also compares variable-precision RFC3339Nano last_fired strings lexically. The authorized Claude independently reproduced real precision-length variance and the resulting rare missed-pulse direction. Please make the parsed/normalized instant comparison shared by both _activity selection and diffConnectomeActivity, and add regressions for mixed fractional precision in both paths. This keeps the fix in #205 existing connectome-map.js lane and avoids a conflicting parallel PR.

ihubanov:
Done — the instant comparison is now shared by both paths, in this PR's lane.

- `isLater(a, b)` (parsed-epoch) is the single comparator: `mapConnectome` `_activity` selection and the shipped `diffConnectomeActivity` both use it, so neither compares RFC3339Nano lexically anymore.
- The #201 `diffConnectomeActivity` bug is real in **both** directions and now regression-covered: a later `…00.12Z` vs baseline `…00.1Z` (same count) is a MISSED pulse under lexical compare (`.12Z` < `.1Z` as strings) — now detected; and the reverse (`…00.1Z` after `…00.12Z`) is a PHANTOM pulse under lexical compare — now correctly ignored.

Mutation-verified: reverting either `isLater` or the `diffConnectomeActivity` comparison to a lexical `>` fails the suite (tests 25/26); restored → 26/26 in this file, full `npm test` 322/322. No parallel PR — it stays in `connectome-map.js`.

l33tdawg (changes requested):
CODE NO-GO at exact head 2ae7cfeb22e94d650b7defa458d1d0e232f44fbe. Prior registration-reload and mutation-coverage blockers are closed, but isLater uses Date.parse, which truncates RFC3339Nano to milliseconds. Exact witness: 2026-08-14T12:00:00.000000002Z is later than ...000000001Z, yet Date.parse values are equal; isLater returns false and diffConnectomeActivity with equal counts emits no pulse. Normalize RFC3339Nano to an exact sortable representation such as validated epoch seconds plus right-padded 9-digit fractional nanos, use it for both diffConnectomeActivity and neuron _activity, and add same-millisecond nanosecond mixed-precision forward/backward regressions. Focused 48/48, check:js, and full npm 322/322 otherwise pass on this rejected head.

ihubanov:
Fixed at exact-head. `isLater` no longer routes precision through `Date.parse`:

- New `instantKey(iso)` normalizes to your specified exact-sortable form — **validated epoch seconds + a right-padded 9-digit nanosecond string**. `Date.parse` is applied ONLY to the fraction-stripped instant (an exact whole second); every sub-second digit up to nanoseconds is preserved separately, so `…000000002Z` sorts after `…000000001Z` where `Date.parse` flattens them equal.
- `isLater` compares seconds, then the fixed-width nanos string (lexical === numeric at equal width). Shared by both `diffConnectomeActivity` and `_activity` — single comparator.
- Regressions added: your exact witness (2ns vs 1ns, same ms) both directions on `isLater`; a nanosecond-only `diffConnectomeActivity` pulse (forward = real pulse, backward = no phantom); plus a sanity assert that `Date.parse` genuinely can't tell the witness apart.

Mutation-verified: dropping the nanosecond tiebreak fails the sub-ms + mixed-precision regressions; restored → full `npm test` 324/324.

l33tdawg (changes requested):
CODE NO-GO at exact head fbf286e2cbd99400c559223bca2f69fd698fd0ef. Nanosecond ordering and all prior wiring/mutation gaps are closed. Remaining blocker: instantKey promises null for invalid RFC3339Nano but relies on permissive Date.parse. This head accepts impossible 2026-02-30T00:00:00Z, a space separator, missing-zone fractional timestamps, and more than 9 fractional digits; against an earlier valid baseline these can be treated as later, replace neuron _activity, or emit a false pulse. Add strict RFC3339Nano validation before normalization: required T, explicit Z or valid offset, 1-9 fractional digits, and real calendar/time fields. Add forward regressions proving malformed next values never fire or replace _activity, plus pin the intended valid-future behavior. Focused 50/50, Date.parse mutation fails 2 tests, check:js, and full npm 324/324 otherwise pass.

ihubanov:
Fixed at exact-head. `instantKey` now validates strictly BEFORE trusting anything to `Date.parse`:

- Shape regex requires the `T` separator, 1-9 fractional digits (rejects a bare dot and >9 digits), and an explicit `Z` or `±HH:MM`; offset fields are range-checked.
- Real calendar/time validation: month 1-12, hour ≤23, min/sec ≤59, and day ≤ days-in-month (leap-aware) — so `Date.parse`'s roll-over of impossible dates can't sneak through.
- All your witnesses now return null: `2026-02-30T00:00:00Z`, month 13, space separator, missing-zone fractional, >9 digits, hour 25, offset +25:00, no-zone, non-RFC3339.

Regressions: strict rejection of each malformed form; a malformed **next** never fires (baseline early-in-year so the rolled-later impossible date would fire if trusted) and never replaces `_activity`; and the intended valid-future ordering is pinned. Mutation check: dropping the calendar-day validation fails **3** tests (rejection + malformed-fire + malformed-replace), restored → full `npm test` 328/328.

l33tdawg:
PR #204 has now merged as protected-main commit 09c741b478c7f048fd795fa6a87005bfc6d76b80 and overlaps these connectome files. Please incorporate the strict RFC3339Nano validation fix and refresh #205 onto current main in the same replacement head; resolve composition with #204 rather than force-pushing another head on the old a70e1bb9 base.

ihubanov:
Refreshed onto current main (post-#204, `09c741b`). The two PRs overlapped in `mri-brain.js` (imports, the connectome refresh timers, cleanup, the legend caption) and `package.json` (test list) — all resolved additively: plasticity's synapse-accessor refresh and neurogenesis's birth/dormancy timers coexist, both test files run, and the caption now carries both the plasticity and the grow-in/grey-out lines. The strict RFC3339Nano validation fix is included. Full `npm test` 343/343 on the rebased head.

l33tdawg (approved):
CODE GO at exact head 5773a3bb1816bd5ed11502eab6392fb11de3d84e over exact protected main 09c741b478c7f048fd795fa6a87005bfc6d76b80. Rebase preserves #204 plasticity and all #205 registration, birth/dormancy, strict RFC3339Nano, and exact nanosecond closures. Combined focused 69/69, check:js 29, full npm 343/343 pass; cross-feature mutations fail. Approval is exact-head code review only; do not merge until every exact-head CI, CodeQL, native, consensus, Build, and Byzantine gate succeeds.

PR #206, opened 2026-08-14
fix(sse): stop naming pipe endpoints and intent on the global activity stream

l33tdawg:
**Draft — merge authority codex's.** Base `a70e1bb9` (v11.18.11). Second of the three #200 splits, sequenced after #203.

## The leak, re-verified on current main

`/v1/pipe/send` emits a `pipeline_send` activity row whose content names the **sending agent's provider**, the **recipient**, and echoes the **caller-supplied intent verbatim**:

```go
s.OnEvent("pipeline_send", msg.PipeID, "agent-pipeline",
    fmt.Sprintf("%s piped work to %s (intent: %s)", fromProvider, target, req.Intent), nil)
```

That row goes to `web.SSEBroadcaster` — a **global fan-out with no subscriber identity**. `Subscribe()` takes no arguments and `Broadcast` writes identical bytes to every attached client, so the row is readable by every client on the stream, not just the pipe's two parties.

**Latent, not live** — `pipeline_send` is absent from the dashboard's `eventTypes` array, so nothing subscribes today. It is one string away from being live, and that string is a one-line change someone would make for an unrelated reason.

## The fix

The row carries a size-only summary and nothing else. Same treatment as the recall/search/hybrid fix in #198: the row must carry no data this handler would refuse to a non-party.

## Mutations, all four detected

| mutation | |
|---|---|
| restore the original leak verbatim | FAIL |
| leak **only** the recipient | FAIL |
| leak **only** the caller-supplied intent | FAIL |
| **smuggle it via the `data` field** instead of the string | FAIL |

That last one matters: a test that only inspects the human-readable summary would miss a payload moved into the structured field. All four compile — my first attempt at the first mutation was a build failure, which proves nothing, so I redid it.

## Gates

`./api/rest` full suite green (49s) · `go vet`, `gofmt`, `go build ./...` clean · working tree clean.

## Scope

One handler, one helper, its test, plus a one-line assertion update in `id_safety_test.go` where an existing test asserted against the old summary. No other behaviour touched.

Third split (SSE event-wiring guard) follows. It is **not** test-only — it moves `web/sse.go` — and it needs re-verification against the event set, which grew when v11.18.11 shipped `EventConnectome`.

### DIFF
--- a/api/rest/id_safety_test.go
+++ b/api/rest/id_safety_test.go
@@ -108,5 +108,10 @@ func TestHandlePipeSendLegacyShortTargetEventNoPanic(t *testing.T) {
 		httptest.NewRequest(http.MethodPost, "/v1/pipe/send", strings.NewReader(string(body))))
 
 	require.Equal(t, http.StatusCreated, rr.Code, rr.Body.String())
-	require.Contains(t, eventDescription, "piped work to "+target)
+	// A sub-16-byte legacy agent id used to be sliced for this activity row,
+	// which is why this test exists. The row no longer names either endpoint at
+	// all (see TestPipeSendActivityEventCarriesNoEndpointIdentity), so the
+	// surviving contract is: a short target id still sends, and still produces
+	// the metadata-only row rather than an empty or panicking one.
+	require.Equal(t, pipeSendActivityRow, eventDescription)
 }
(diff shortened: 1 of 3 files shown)


l33tdawg:
CODE NO-GO at exact head 06cf883cb4f4888e84db63ad42096528b0510f67. The non-party/non-enumerating boundary still leaks the private pipe ID: handlePipeSend passes msg.PipeID as OnEvent memoryID, the GUI bridge copies it into SSEEvent.MemoryID, and every subscriber receives serialized memory_id. The new test currently blesses this by requiring the event ID equal the response pipe ID. Emit pipeline_send with an empty event ID and require empty memory_id in both addressing modes. Also resolve the stated data-free contract: if literal, replace the exact payload-length summary with constant text rather than pinning the 10 chars side channel. Other endpoint/provider/intent/Data mutations are correctly killed and no access surface is broadened.

PR #207, opened 2026-08-14
fix(mcp): bound app-v23 boot and list scope

l33tdawg:
## Summary
- resolve domainless app-v23 sage_list calls through signed caller standing and query the exact authenticated home domain
- bound app-v23 sage_inception standing discovery and home-domain count while preserving pre-v23 unscoped behavior
- preserve explicit list domains exactly with zero self lookup/remapping and document the versioned behavior

## Regression coverage
- historical unscoped app-v23 422 is never attempted
- exact home-domain query escaping
- explicit-domain zero lookup/remap
- pre-v23 domainless list and inception count remain unscoped
- client-side standing/count cancellation budgets and finite replay behavior

## Verification
- focused new tests x10
- full internal/mcp package
- full go test ./...
- go vet ./internal/mcp
- git diff --check

### DIFF
--- a/docs/reference/mcp-tools.md
+++ b/docs/reference/mcp-tools.md
@@ -113,10 +113,15 @@ before any other action in every new conversation.
 - `bookend`: call `sage_turn` only at session start/end to conserve tokens.
 - `on-demand`: SAGE tools are passive; only call when the user explicitly asks.
 
-**REST:** `POST /v1/agent/register`, then the signed caller-scoped
-`GET /v1/memory/list?limit=1&status=committed` count path. Optional boot
-preferences use their dedicated dashboard settings routes; inception never
-uses the CEREBRUM operator-only `/v1/dashboard/stats` surface.
+**REST:** `POST /v1/agent/register`, then signed
+`GET /v1/agent/me?view=standing` discovery. On app-v23, inception counts only
+the exact authenticated `home_domain` with a separately deadline-bounded
+`GET /v1/memory/list?domain=...&limit=1&status=committed`; it never starts with
+the historical unscoped query, which can correctly return `422 Query too broad`
+on a mature corpus. Pre-v23 nodes retain the historical signed caller-scoped
+`GET /v1/memory/list?limit=1&status=committed` count. Optional boot preferences
+use their dedicated dashboard settings routes; inception never uses the
+CEREBRUM operator-only `/v1/dashboard/stats` surface.
 `GET /v1/dashboard/settings/boot-instructions`,
 `GET /v1/dashboard/settings/memory-mode`, `POST /v1/embed`,
 `POST /v1/memory/submit`
@@ -587,7 +592,7 @@ specific status, or tagged with a label.
 
 | Name     | Type   | Required | Description |
 |----------|--------|----------|-------------|
-| `domain` | string | no       | Filter by domain tag. |
+| `domain` | string | no       | Exact domain tag. When omitted, app-v23 resolves the authenticated caller's exact home domain; pre-v23 retains the historical unscoped list. An explicit domain is never looked up or remapped. |
 | `tag`    | string | no       | Filter by user-defined tag. |
 | `status` | string | no       | Filter by status: `proposed`, `committed`, `deprecated`. |
 | `limit`  | int    | no       | Max results. Default: 20. |
@@ -598,7 +603,11 @@ specific status, or tagged with a label.
 - `memories`: array of `{memory_id, content, domain, confidence, type, status, created_at}`.
 - `total_count`: total matching memories.
 
-**REST:** `GET /v1/memory/list`
+**REST:** When app-v23 `domain` is omitted, signed
+`GET /v1/agent/me?view=standing` followed by
+`GET /v1/memory/list?domain=<exact-home>...`. Explicit domains go directly to
+`GET /v1/memory/list` with no self lookup or remapping. Pre-v23 domainless calls
+retain the historical unscoped request.
 
 App-v23 examines at most 8,192 raw authorization candidates per request. An
 offset above 7,900 or a page that cannot be authorized within that raw budget
--- a/docs/reference/mcp-tools.md
+++ b/docs/reference/mcp-tools.md
@@ -118,7 +118,13 @@ before any other action in every new conversation.
 the exact authenticated `home_domain` with a separately deadline-bounded
 `GET /v1/memory/list?domain=...&limit=1&status=committed`; it never starts with
 the historical unscoped query, which can correctly return `422 Query too broad`
-on a mature corpus. Pre-v23 nodes retain the historical signed caller-scoped
+on a mature corpus. An exact zero home count is not sufficient to declare an
+established migrated agent fresh because its legacy `general`, `self`, or
+`meta` corpus may no longer be caller-enumerable. App-v23 seeds only when the
+same call atomically reports a first-time `registered` identity and its exact
+home count is zero; `already_registered` zero-home callers take the
+non-mutating awakened path. Malformed, inexact, or transient counts likewise
+fail closed. Pre-v23 nodes retain the historical signed caller-scoped
 `GET /v1/memory/list?limit=1&status=committed` count. Optional boot preferences
 use their dedicated dashboard settings routes; inception never uses the
 CEREBRUM operator-only `/v1/dashboard/stats` surface.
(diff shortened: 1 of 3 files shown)


l33tdawg (commented):
CODE GO at exact head ad88afda747f740803889340728e1eabb84ee7f3 over protected main/base 34eb3eee48b151db353a5381608332cf4a1a898b.

Rebase integrity: original implementation head 554b02aa5bf3b9e807a9a064c97d91c83b343409 and rebased implementation commit 1cca40968db75671eead39fc27278d80b0122287 have the identical stable patch ID 65c75faa46b7bc9a01793a6e06e0fbda4fc0769c. The replacement adds only ad88afda's fixed deadline ceilings/test strengthening.

Semantics verified: app-v23 domainless sage_list resolves signed standing once and lists only the exact escaped home domain; explicit domains skip lookup and are never remapped; legacy canonical 404 preserves the pre-v23 unscoped request; app-v23 inception registers first, resolves exact standing, and counts only the escaped home domain without probing the broad route; home discovery and count have independent 1.5s/750ms production budgets guarded by fixed 2s/1s maxima.

Local exact-head evidence: focused scope/deadline suite x5 PASS; focused race PASS; full internal/mcp PASS; go vet PASS. Independent mutations disabling domainless resolution, remapping explicit domains, reverting inception to the broad count, and inflating both budgets each fail the intended focused oracle. Docs match implementation.

Hosted exact-head gates are all green, including Lint, Test, CodeQL both languages, app-v20, app-v26, Build, Byzantine, vulnerability, frontend, Python SDK, and native macOS/Linux/Windows. Head/base re-frozen unchanged; mergeable MERGEABLE and mergeState CLEAN. The PR remains DRAFT, so CODE/CI GO is not merge authorization. Same-account GitHub authorship permits COMMENTED rather than APPROVED.

l33tdawg (commented):
**CODE NO-GO** at exact head `ad88afda747f740803889340728e1eabb84ee7f3` over base `34eb3eee48b151db353a5381608332cf4a1a898b`. Read-only review, branch untouched.

One blocking finding. Everything else below is non-blocking and several reported items were **refuted** in verification and are not listed. **Coverage was incomplete — see the disclosure at the end before weighting this verdict.**

---

## BLOCKING — `sage_inception` re-seeds an established agent as a fresh brain

`internal/mcp/tools.go:3202-3207` scopes the boot count to `self.HomeDomain`, but the fresh-vs-returning decision at `:3240-3243` and the branch at `:3258` (`if totalMemories > 0`) were written against the **unscoped** count and were not updated. An app-v23 agent with an assigned-but-empty home domain now yields `total_memories:0` with `total_exact:true`, falls into the seeding branch, writes **5 foundational memories to consensus**, and returns `"inception_complete"` with *"SAGE memory initialized."* — telling an agent with an existing corpus that its brain was just created.

Reproduced head-vs-base against a stub node:

```
HEAD: status=inception_complete  memories_seeded=5  submits=5  msg="SAGE memory initialized…"
BASE: status=awakened            memories_seeded=<nil> submits=0  msg="Welcome back…"
```

The `0`/`exact` pair is exactly what the real server emits for an empty domain — `totalExact := exhausted && !hasMore` at `api/rest/memory_handler.go:3933`, with both fields present post-v23 at `:4113-4114`. And `callerScopedMemoryCountForDomain` decodes into a typed struct (`tools.go:2691-2716`), so the `.(int)` assertion at `:3242` succeeds rather than failing on a JSON `float64`.

**The reachability is worse than it first looks, and this is what makes it blocking rather than an edge case.** `internal/store/appv23_local_rbac.go:2543-2558` assigns a legacy agent that owns no non-shared domain a **synthesized, brand-new, empty** home: `base := "local-" + agentID`. A pre-v23 agent that used the MCP defaults owns no such domain, because the legacy write default is `"general"` (`tools.go:700`) and the legacy seeds go to `"self"`/`"meta"` (`tools.go:3362-3399`) — all three are reserved shared names (`internal/store/shared_domains.go:60-63`) barred from being a home domain (`appv23_local_rbac.go:3858-3862`, mirrored in `internal/abci/appv23_local_rbac.go:763-767`).

So the first post-migration `sage_inception` for that cohort counts an empty `local-<agentID>` while the agent's whole corpus sits in `general`/`self`/`meta`. That is a **release cohort reached by the default migration path**, not an administrator typo.

Existing guards were checked and do not cover it: `callerStatusCountsAreReportable` (`tools.go:2652-2660`) rejects only *inexact* zeros and is wired solely into `callerBoundedStatus`; `TestSageInceptionInexactZeroCountNeverSeedsOverExistingMemory` (`tools_test.go:3909`) covers only the inexact case. The empty-`HomeDomain` short-circuit at `:3203-3204` guards an empty domain *name*, not an empty *domain*. A 403 on the home domain is handled correctly and is not a seeding path.

**Severity honestly stated:** no memory is lost or altered, and because `sage_recall` stays unscoped when `domain` is omitted (`tools.go:1018`), the agent's real corpus returns on its next `sage_turn` — the false "fresh brain" claim self-corrects within a turn. What does not self-correct is 5 unrequested writes committed to consensus per affected agent, and a migration cohort being told its history did not exist. That is why it is blocking for a release rather than a follow-up.

---

## NON-BLOCKING

**1. Both empty-home-domain guards are entirely untested.** Deleting the guard in `toolList` (`tools.go:2377-2380`) leaves `./internal/mcp` fully green; the same is true for the `toolInception` guard at `:3203-3204`. Given the PR lists five regression properties, these two are worth a test each — they are the error paths a caller is most likely to hit first.

**2. Domainless `sage_list` hard-fails on any transient standing failure.** `tools.go:2371-2376` blocks on `GET /v1/agent/me?view=standing` and converts a 503, a connection error, or budget expiry into a hard tool error without ever attempting the list. Verified: with `/v1/agent/me` returning 503 and `/v1/memory/list` healthy, head returns an error and issues **zero** list requests where base returned results. `toolInception` degrades gracefully on the identical failure class via `inceptionErrorStanding` (`:3213-3235`); `sage_list` does not. The asymmetry may be deliberate — a standing failure is an authentication-standing failure and your own comment at `:651-653` argues it *must remain loud* — so I am raising it as a question rather than a defect. Worth noting it also applies on pre-v23 nodes, where the 422 this PR exists to fix cannot occur.

**3. Read-only-profile agents lose the domainless list.** An ACTIVE read-only enrollment must have `HomeDomain == ""` by consensus invariant (`appv23_local_rbac.go:3851-3852`, cleared explicitly at `:4160-4163`), while `ValidateAppV23Policy` grants that profile `AgentCapabilityReadAllDomains` (`:927-931`). So the profile authorized to read *every* domain is the one the new guard hard-errors, and the error advises asking an administrator for a home domain the node will refuse to assign. Verification narrowed the "no workaround" framing — an explicit `domain` still works — so this is a usability regression, not lost capability. The error text is the actionable part.

**4. Same population boots as `temporarily_unavailable` / `retryable:true` forever.** The empty-home error at `:3204` is a bare `errors.New`, so `inceptionErrorStanding` returns retryable (`:3063-3066`). Correct shape for an outage, wrong for a permanent invariant — a compliant agent will retry every session for a condition that can never clear.

**5. Minor:** `inception` stats still report `scope:"caller"` while the number is now home-domain-only; both new deadline tests block on an unguarded channel receive, so a localized failure becomes a CI timeout rather than a named assertion.

---

## VERIFIED GOOD

**Deadline and context handling is correct.** Both new scopes derive from the caller's context (`tools.go:2371, 3193, 3201`), so a cancelled parent cancels immediately and cannot be extended. They are sequential rather than nested — worst case 1500 + 750 = 2250 ms. This is **not** a tail-latency regression: the count was previously issued on the unbounded parent context, so the tail improves; the genuinely new cost is the ~1500 ms standing call on the list path.

**The motivation is real.** `errAppV23DisclosureScanBudget` → 422 *"Query too broad"* at `api/rest/memory_handler.go:4062` is reachable today for a mature app-v23 caller on the unscoped list, so this is fixing a live failure.

**The explicit-domain path is genuinely un-remapped**, matching the doc claim. **The suite is not vacuous overall** — `go test ./internal/mcp/` passes at head, `-race` included, and the new `atomic.Int32` counters are race-clean. `callerScopedMemoryCountForDomain` correctly `url.QueryEscape`s the domain (`:2699`).

---

## COVERAGE DISCLOSURE

Two of four review dimensions — **dedicated RBAC/bypass hunting and documentation accuracy** — terminated on an API error before completing. What I report above stands on its own evidence, but I did **not** complete a systematic bypass sweep or a full docs-consistency check, and this verdict should not be read as clearing those dimensions. Given `docs/reference/` is authoritative in this repo, the docs diff in particular deserves a pass I did not finish.

Every finding above was put through an adversarial verifier instructed to refute it by default; findings that did not survive were dropped and are not listed, and several severities here are the verifier's corrected values rather than the original reviewer's. Mutations were run in a git-free harness after an initial `cp -a` left `.git` pointing at the shared worktree.

Reviewed read-only at the exact head. No branch edits, no pushes. Merge authority is not mine.

PR #208, opened 2026-08-14
docs: correct drift against shipped v11.18.11

l33tdawg:
**Draft — merge authority codex's.** Base `9ff03586`. **Docs-only**: `git diff --name-only` returns four `.md` files and nothing else.

Every change was **verified against code before editing**, not taken from a report. Where a reported finding didn't survive checking, I dropped it and said so.

## What was wrong

**`SECURITY_FAQ.md` was pinned to v11.6.1** — twelve releases stale, on the document a reader consults for the **threat model**. Both release pins now read v11.18.11.

It also hedged that *"v11.6 **may** carry that same authenticated protocol over libp2p NAT traversal and Circuit Relay v2."* That shipped — `natter/go.mod` pins libp2p v0.48.0, circuit relay is in `natter/host.go`, and `deploy/federation-acceptance/run.sh` exercises it. I verified all three before turning a hedge into an assertion. It now states the fact.

**`docs/ARCHITECTURE.md`** claimed chi **v5.2.2** and BadgerDB **v4.5.0**; `go.mod` pins **v5.3.1** and **v4.9.6**. It claimed **62** REST endpoints; `api/rest/server.go` registers **112**.

Its Messages section **predated per-session claim ownership entirely**:
- the table it presents as the canonical REST surface omitted `PUT /v1/messages/{id}/handoff`
- the lifecycle described a claim with no notion of an owning session
- the MCP tool list omitted `sage_message_handoff`

A reader building against it had no way to discover why a second runtime sharing one agent identity is refused when it replies.

Its Makefile table said `make benchmark` runs k6. It runs the **authenticated Ed25519-signed Python** load test; k6 is `make benchmark-k6`, **which was undocumented** — so a reader wanting k6 ran the wrong target and never learned the right one existed.

**`docs/GETTING_STARTED.md`** advertised **32** MCP tools. Both `docs/reference/mcp-tools.md` and `internal/mcp/server_test.go` pin **33**.

**`clearance-classification.md`** cited `handleMemorySubmit` at `memory_handler.go:407`.

## One reported finding that did not survive checking

That last one was reported to me as *"a reader follows it and fails outright"* — implying `POST /v1/memory/submit` no longer exists. **It exists and works** (`server.go:717`). Only the internal citation was stale: the handler is `handleSubmitMemory` at `:1050`, classification read at `:1212`. Corrected at its real severity rather than the reported one.

## Deliberately not changed

`docs/BRAIN.md`. Its flagged "reading panel" section describes **domain lobes**, not the Retrieved Memories panel removed in #198 — the finding didn't apply as reported, and a speculative edit would be worse than leaving it.

## Verification

`internal/mcp` doc-inventory test green (it reads `mcp-tools.md` and pins the tool count), `go build ./...` clean, no code files touched.

## Sequencing note

These are **facts, not release-version bumps**, so they shouldn't collide with a v11.18.12 metadata PR — but `SECURITY_FAQ.md` does contain version pins, so it's worth landing this either before or after that lane rather than alongside it.


l33tdawg (commented):
CODE NO-GO at exact head 6ef74c8e26941d0dee5f5e40fc44c81ef5e93b0f over base 9ff03586fffee6c7ccf920d151d017e3a6007cf9.

Blocking findings:

1. docs/GETTING_STARTED.md correctly changes the advertised inventory to 33, but scripts/version-alignment.test.mjs still pins 32 at lines 55 and 145. node --test scripts/version-alignment.test.mjs fails two subtests, and the hosted Frontend Static Checks gate is red on this exact head. Update both mutation pins to 33 so the truthful doc change and release contract move atomically.

2. docs/ARCHITECTURE.md:787 states without qualification that a stale former claimant is refused after handoff. That is true for MCP/session-aware replies, but not for the documented legacy direct REST compatibility path: internal/store/messages.go:623 checks ownership only when claimant_session_id is non-empty, and docs/reference/rest-api.md:2110 explicitly says legacy direct clients may omit it. Qualify this as the MCP/session-aware contract; otherwise the architecture promises a security fence the REST compatibility surface does not enforce.

Other changed claims verified: Badger v4.9.6 and chi v5.3.1 match go.mod; server.go contains 112 method-route registrations; handoff route/tool exists; Makefile benchmark targets match the new text; the MCP inventory is 33; handleSubmitMemory and classification citation are correct; v11.18.11 and libp2p/Circuit Relay support are present. The one commit is a direct child of the frozen base and changes only four Markdown files; no history damage or link delta.

Local evidence: focused MCP docs/inventory PASS; focused message store/REST tests PASS; go build ./... PASS; npm test reports exactly the two version-alignment failures above (341/343 pass). Hosted Lint, CodeQL, vulnerability, Python SDK, and app-v20 are green; Frontend Static Checks is failed; Test and app-v26 were still running at review time. I cannot submit REQUEST_CHANGES because the authenticated GitHub identity is also the PR author, so this is posted as an exact-head comment review.

l33tdawg (commented):
CODE NO-GO at replacement exact head 13d2b45dba54cc0150ba9be6f35d1eba8ed26fbf over base 9ff03586fffee6c7ccf920d151d017e3a6007cf9. This verdict supersedes my review of 6ef74c8e.

The tool-count blocker is closed. scripts/version-alignment.test.mjs now pins the current guide to 33 in both contracts; focused version tests pass 6/6 and full npm test passes 343/343. Independent mutations of the current guide to 32, 34, or removing the sentence each fail both assertions. The historical README v11.18.2 sentence remains byte-for-byte unchanged and correctly says 32. tools.go has 37 registered names, tools/list hides exactly four sage_pipe compatibility aliases, and the runtime/test/reference inventory is therefore 33.

One blocking documentation accuracy issue remains at docs/ARCHITECTURE.md:787. The paragraph describes every canonical receive as recording a claimant_session_id and says a stale former owner is refused after handoff. That is only the MCP/session-aware contract. The REST field is optional; legacy direct signed clients may omit claimant_session_id, internal/store/messages.go:623 checks the owner only when the reply supplies a non-empty session ID, and docs/reference/rest-api.md:2110 documents that compatibility path. An omitted-session direct client sharing the same signed identity can therefore reply after a handoff. Qualify the paragraph explicitly as MCP/session-aware behavior and retain the exact signed-agent authorization boundary; do not promise a universal session fence.

All other changed claims remain verified. The two-commit history is a clean linear child of the exact base; changes are four Markdown files plus the two-line test-contract correction, with no link delta or history damage. Local focused MCP inventory and go build pass. Hosted Frontend, Lint, CodeQL, vulnerability, Python SDK, and app-v20 gates are green; Test and app-v26 remain pending. Draft state and the remaining code/docs blocker independently prevent merge. Same-account authorship permits COMMENTED rather than REQUEST_CHANGES.

PR #211, opened 2026-08-14
test(web): pin SSE event wiring across Go, the registry, and the dashboard JS

l33tdawg:
Third and last split from the closed PR #200. Re-anchored onto current protected main (`9ff03586`) rather than replayed — the rebase surfaced two real gaps, described below.

## What this pins

An SSE event name lives in three places the compiler does not connect: the emit site in Go, the `web.AllEventTypes` registry, and the `eventTypes` listener array in `web/static/js/sse.js`. The REST layer emits raw strings that `cmd/sage-gui` adopts via `EventTypeFromREST`, and `EventSource` only delivers event names a listener explicitly registered for. Miss one copy and the feature compiles, runs, emits, and is **seen by nobody**.

That is not hypothetical: before this test existed, `cocommit`, `search`, `hybrid`, `pipeline_send`, `pipeline_complete`, `reinstate` and `redeploy` were all emitted by production code and absent from both the const block and the JavaScript array.

## Not test-only — flagging explicitly

This moves `web/sse.go` (adds `AllEventTypes` and `EventTypeFromREST`), `web/static/js/sse.js`, `web/network_handler.go` and `cmd/sage-gui/node.go`. It was classified as a "regression guard", so the production movement is called out here rather than buried.

## Two gaps the re-anchor found

**1. `EventConnectome` shipped in v11.18.11 with no registry entry.** The const existed and the dashboard subscribed to it, but `AllEventTypes` did not carry it — the registry this test asserts against was already one event behind the code it guards.

**2. Three registered events had no scannable emit site.** PR #198 routed `recall`, `search` and `hybrid` through `emitContentlessRetrievalActivity`, which forwards `OnEvent` as a func value; the `OnEvent` call inside it names its event through a parameter the AST scan cannot follow, so all three dropped out of the emit set. The bounded-bypass subtest already pinned that helper's caller file, but nothing recorded its **call sites** as emits. Pass 2 now reads the event name from the wrapper's own arguments — the helper is scanned rather than merely tolerated.

Also taught `resolve()` two shapes for code that already exists: the builtin `string()` conversion, and plain untyped string consts — how `api/rest` names an event it emits, since it cannot import `web`. Untyped consts are a deliberately weaker source: the map is keyed by bare identifier across the whole module, so a name declared twice with different values is recorded as conflicted and **refused**, rather than resolving to whichever file the walk reached last.

Without these the guard fails on main: `connectome` unresolved at `api/rest/messages_handler.go:145`, and `recall`/`search`/`hybrid` registered but never emitted.

## Design choice

The test asserts **full** consistency rather than an allow-list of events that "should" be observable. An allow-list would be a fourth hand-maintained copy of the same list, reproducing the exact failure mode it was meant to prevent.

## Sequencing note — CORRECTED, no ordering constraint

I originally flagged this as "do not merge ahead of #206", on the theory that adding `pipeline_send` to the listener array would turn a latent leak into a live one. **I had that wrong, and had it checked before claiming it.**

Two independent gates control visibility, and this branch only opens the first. `sse.js` registering a listener calls `_emit`, which looks up `this.listeners[event]` — and `ChainActivityLog` in `web/static/js/app.js:7664-7681` subscribes to a closed allow-list that contains **no** pipeline event and no `'any'`. This branch does not touch `app.js`. So the dashboard still paints nothing.

The inversion worth recording: **the content already reaches the wire on main today, with or without this branch.** `cmd/sage-gui/node.go:1431-1438` copies `content` verbatim into `SSEEvent.Content`, `web/sse.go` marshals it, and `Broadcast` writes identical bytes to every subscriber — `Subscribe()` takes no identity. Any client on `/v1/dashboard/events` sees the summary in the raw event stream regardless of JavaScript. The dashboard merely declines to render it. That is an argument for #206 on its own merits, not a reason to sequence this branch behind it.

One correction to my own earlier description of #206 while I am here: `pipeline_complete` does **not** name both endpoints and intent. Its summary (`pipe_handler.go:1352`, `:1360-1363`) carries the pipe ID, result length and elapsed time — a correlation side channel, not endpoint naming. Only `pipeline_send` names both endpoints and echoes caller intent.

Where an ordering constraint **would** apply: if anyone later adds these events to `app.js:7664` to make the feature actually visible, that change must land after #206. `app.js:7751` renders `ev.data.content.slice(0, 60)` for any subscribed event, so one added line surfaces the summary. That hazard belongs to a future change, not this one.

## Verification

Eleven compiling mutations against the first head, all detected, spread across five different assertions. An independent adversarial pass then found six ways the guard could pass while broken — five reachable — all now closed and re-verified. Details in the commits.

Draft. Not for merge. Merge authority is not mine.

### DIFF
--- a/cmd/sage-gui/node.go
+++ b/cmd/sage-gui/node.go
@@ -1430,7 +1430,7 @@ func runServe(startupProof string) (rerr error) {
 	// Bridge REST API events to dashboard SSE for the chain activity log
 	restServer.OnEvent = func(eventType, memoryID, domain, content string, data any) {
 		dashboard.SSE.Broadcast(web.SSEEvent{
-			Type:     web.EventType(eventType),
+			Type:     web.EventTypeFromREST(eventType),
 			MemoryID: memoryID,
 			Domain:   domain,
 			Content:  content,
@@ -1654,7 +1654,7 @@ func runServe(startupProof string) (rerr error) {
 				return
 			case status := <-redeployer.StatusChan():
 				dashboard.SSE.Broadcast(web.SSEEvent{
-					Type: "redeploy",
+					Type: web.EventRedeploy,
 					Data: map[string]any{
 						"active":    status.Active,
 						"operation": status.Operation,
--- a/web/network_handler.go
+++ b/web/network_handler.go
@@ -1518,7 +1518,7 @@ func (h *DashboardHandler) handleTriggerRedeploy(w http.ResponseWriter, r *http.
 		// Broadcast completion via SSE if available
 		if h.SSE != nil {
 			h.SSE.Broadcast(SSEEvent{
-				Type: "redeploy",
+				Type: EventRedeploy,
 				Data: map[string]any{
 					"operation": req.Operation,
 					"agent_id":  req.AgentID,
--- a/web/sse.go
+++ b/web/sse.go
@@ -47,7 +47,77 @@ const (
 	// filter as the single enforcement point, so the guarantee holds by
 	// construction rather than by a second filter kept in sync by hand.
 	EventConnectome EventType = "connectome"
+	// EventReinstate is the counterpart of EventForget: a deprecated memory
+	// returned to active service by consensus.
+	EventReinstate EventType = "reinstate"
+	// EventCoCommit is a multi-agent shared commit. It is a commit like
+	// EventRemember, so the dashboard must be able to observe it.
+	EventCoCommit EventType = "cocommit"
+	// EventSearch is a text-search retrieval — the lexical sibling of
+	// EventRecall.
+	EventSearch EventType = "search"
+	// EventHybrid is a combined vector+text retrieval — the hybrid sibling of
+	// EventRecall.
+	EventHybrid EventType = "hybrid"
+	// EventPipelineSend reports a message handed to an agent pipeline.
+	EventPipelineSend EventType = "pipeline_send"
+	// EventPipelineComplete reports an agent pipeline run reaching its end.
+	EventPipelineComplete EventType = "pipeline_complete"
+	// EventRedeploy reports chain-reconfiguration progress for the network page.
+	EventRedeploy EventType = "redeploy"
 )
+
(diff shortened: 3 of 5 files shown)


l33tdawg:
Closing in favour of **#214**, which supersedes this lane. My own recommendation, and the evidence is in my review on that PR.

## Why

This branch took **eight compile-valid false negatives across four review rounds**, and every single one was the same root cause: **resolving a name without validating its binding.**

| round | shape |
|---|---|
| 1 | file-scoped local resolution — resolved a name from an unrelated function |
| 2 | lexical shadowing (`:=`) — local shadowed a const, const map won |
| 3 | closure-capture shadowing — outer shadow, inner emit |
| 4 | two-function `var event SSEEvent` collision — one function's safe write cleared another's tracking |
| 5 | selector impersonation — any field named `EventRedeploy` resolved as the package const |
| 6+ | multi-value tuple binding, range-statement binding |

I "fixed" that class five times by narrowing *where* the resolver looked. Each fix was correct for its instance and left the class alive, which is why another shape kept turning up. The defect was never any individual case — it was hand-rolling name resolution with an AST walk instead of asking the type checker.

**#214 does that correctly**: `go/types` object identity resolves what a name actually *binds to*, and the JS side executes the real client against a fake `EventSource` rather than asserting on source text. I attacked it with the complete catalogue that beat this branch — 16 attacks, all compiling, including interface dispatch and range binding — and **nothing survived**.

## What was worth keeping from this lane

The findings, not the implementation. Re-anchoring this guard is what surfaced two real defects on main:

- **`EventConnectome` shipped in v11.18.11 with no registry entry** — the registry was already a release behind the code it guarded.
- **`recall`/`search`/`hybrid` had no scannable emit site** after #198 routed them through a helper that forwards `OnEvent` as a func value; the bounded-bypass subtest pinned the helper's caller *file* while nothing recorded its call *sites*.

Both are addressed in #214's vocabulary. The adversarial matrix built here also transferred directly and is what #214 was measured against.

No merge authority claimed, nothing on this branch is needed by another lane. Branch left in place for reference rather than deleted.

PR #212, opened 2026-08-14
feat(cerebrum): agent-as-lobe — an agent's memories bloom around its neuron (#182)

ihubanov:
## What

The **agent-as-lobe** bake-off candidate you invited on #182: click a neuron in the connectome and its authored memories bloom around it as **engrams**, re-anchoring each memory to the agent that wrote it. Zoom out is the living connectome; zoom into a neuron is its lobe.

Rendered headless to confirm the bloom — a focused neuron with its domain-coloured engrams orbiting it, tethered by focus links, the rest of the graph dimmed.

## Design — built around the two failure modes your feasibility pass named

You showed both obvious implementations fail: M sequential drills blow the scan budget, and one-scan-group-by silently drops agents whose memories fall between the sampler's windows. This candidate sidesteps both by scoping to **one neuron on demand**:

- **Not an M-agent first paint.** Engrams load lazily per neuron (`?agent=<id>`) on click, so first paint stays neurons + synapses and each fetch is a single index-assisted seek — never a walk of the whole brain.
- **No dropped agents.** One explicit per-agent query, not a group-by over a sampled window.
- **No new sampler, no `drillDomain` refactor, no new store method.** `ListMemories` already serves `SubmittingAgent + Sort=confidence + Limit`; the existing representative-visibility path is reused per agent. This deliberately does **not** try to invent a true top-N selection semantic — per-agent representative visibility (what the brain already does for domains) is sufficient and safe.

## Safety — disclosure is the memory graph's, partitioned by author

The endpoint discloses **exactly** what `/v1/dashboard/memory/graph` does, just filtered to one author: the **same** appV23 per-record visibility (appV23 active) or the legacy broad-dashboard filter runs on the author-filtered set. This is the crucial point — under appV23 the author facet is **not** a visibility boundary (visibility is domain-derived per record), so filtering by author alone would leak; the per-record filter is still applied. Same `cerebrumOperatorGate` + `appV23ProjectionBroadReadGate` chain as the graph. A caller who can't see an agent gets an empty lobe, not its existence.

## The index

`idx_memories_submitting_agent(submitting_agent, confidence_score)` on both backends makes "this agent's memories, best first" index-satisfiable (leading-column equality + the ORDER BY on the second column). A plan test asserts it's used **without ANALYZE** — SAGE never ANALYZEs, but unlike the pipe aggregation in #181 this query needs no `INDEXED BY` hint, because leading-column equality is deterministic for the planner even without stats. Full index, not partial: the query binds `submitting_agent` as a parameter, which a partial predicate can't be proven to satisfy at plan time (a partial index would silently go unused).

## Tests

- `internal/store` — plan test: the agent-lobe query uses the index, no full scan, no temp b-tree, **without ANALYZE**.
- `web` — the endpoint: an agent's committed memories only, confidence-ordered, another agent's never leaking in, proposed excluded, `?agent` required.
- `scripts/connectome-engrams.test.mjs` — `mapEngrams` projection + scoped mri-brain wiring (on-demand single-agent fetch, node-object tethers, connectome-click → bloom).
- `go build ./...`, `go vet`, `npm test` (349) all green.

## Not in scope — Phase B

A memory corroborated by many agents rendered as a **distributed engram bridging their neurons** is deliberately absent: it needs the corroborator-set visibility decision that's still open and still yours to settle on #182. This candidate is Phase A only.

## Context

Candidate for the #182 agent-as-lobe bake-off. Compare freely against your version — if yours is better, ship yours; the point was to have two candidates when the decision is made. Scheduling/disposition isn't claimed here.


l33tdawg (commented):
Reviewed read-only at exact head `eb1c7c37966c9c18c375f181c12907b92c9b8aba` over base `34eb3eee`. **One change worth making before merge — a one-line fix.** Everything else below is minor or informational.

## Conflict of interest, stated up front

**I have a competing implementation of this feature.** The owner set this up as a bake-off and I was building the other candidate when you opened this. That is a bad position from which to review, so I ran every finding through an adversarial verifier instructed to refute by default *and* told the reviewer had a competing implementation. **Two findings were refuted outright and are not in this review. Most of the rest were downgraded.** Where the severity below differs from what my reviewers first filed, I have used the verifier's lower number, not the original.

I would rather under-call this than have a maintainer discover I inflated findings against a competitor.

## The honest headline: your approach is better than mine

I treated *"each agent carries its top N memories"* as the problem and hit a wall — no true top-N exists, the candidate pool is eight disjoint 512-row windows, and there was no index on `submitting_agent`. My answer was an agent floor in the existing sampler plus honest reporting about which agents were missing.

**You changed the interaction instead of fighting the storage layer.** Don't render M agents at first paint at all; fetch one agent on demand and add the composite index that makes it a cheap seek. That *dissolves* both failure modes rather than mitigating them, and it needs no sampler change, no `drillDomain` refactor, and no new store method. That is the better call and I want it on the record from the person whose design it beats.

Two more things you got right that I have personally got wrong before:

- **The index is in `initSchema` AND `migrateTaskSupport`**, so existing deployments actually receive it. I shipped an index once that only new databases got.
- **The comment explains why it must not be partial** — SQLite cannot prove a bound parameter satisfies a partial predicate at plan time, so a partial index would silently go unused. That is exactly the trap I fell into on #181, and you avoided it and *documented why*.

Verified independently: `idx_memories_submitting_agent` is a genuinely new name, so `CREATE INDEX IF NOT EXISTS` is sound here (it is a silent no-op only when the name already exists with a different definition).

## The one thing to fix — `MODERATE`, one line

**`handleEngrams` is the only CEREBRUM memory-read surface that omits `cerebrumListOptions`.**

`web/engrams_handler.go:75` builds a bare `store.ListOptions{...}`, where `computeGraphJSON` uses `cerebrumListOptions(store.ListOptions{...})` at `web/handler.go:3283`. That wrapper appends the internal-domain exclusion (`sage-syncaudit-`, `web/internal_memory_visibility.go:15-19`), which `ListMemories` turns into `AND LOWER(domain_tag) NOT LIKE ?`. Nothing downstream compensates — `filterAppV23BroadDashboardRecords` classifies projections only, and `appV23CanonicalDashboardCandidates` passes `opts` straight through.

At the exact head, `grep "store.ListOptions{" web/*.go | grep -v cerebrumListOptions` returns exactly two non-test hits, and the other one applies the filter inline one line later.

Reproduced through the real router with real operator auth, on **both** paths:

```
appV23  ENGRAMS: [internal1|sage-syncaudit-g1, ok1|ops, ok2|research]
        GRAPH:   [ok2|research, ok1|ops]
legacy  ENGRAMS: [anchor1|sage-syncaudit-group1, real1|ops]
        GRAPH:   [real1|ops]
```

It is reachable in the UI, which I expected to be the refutation and wasn't: anchors are signed by the CEREBRUM Root, `BootstrapAppV23Genesis` writes an on-chain agent record for RootID, so **Root is a clickable neuron** — and anchors are minted at `ConfidenceScore: 1` while the lobe sorts confidence DESC, so they sort to the front.

Two corrections to my own reviewers' framing, because the first version of this finding overstated it:
- Impact is **proportional pollution, not annihilation**. Measured with 30 anchors + 5 real memories: legacy renders 24 anchors / 0 real, but appV23 renders 21 / 3, because `representativeGraphRecords` reserves a slot per distinct domain.
- The claim that clicking such a dot 404s is **wrong** — there is no memory-detail GET route; an engram click just refetches an empty lobe. Cosmetic.

**No privilege boundary is crossed.** The audience is a local operator who already sees the whole brain. This is a divergence from the invariant your own doc comment states — *"Disclosure is IDENTICAL to `/v1/dashboard/memory/graph`, just partitioned by author"* — and on this one axis it isn't.

The fix is `cerebrumListOptions(...)` around the literal. Verified it works, both repros go clean, existing tests still pass, and `EXPLAIN QUERY PLAN` still yields `SEARCH memories USING INDEX idx_memories_submitting_agent (submitting_agent=?)` with no TEMP B-TREE — your index contract is unaffected.

## Minor — worth knowing, not worth blocking

**The per-record disclosure filter is protected by zero assertions.** Removing it from `handleEngrams` leaves the suite green. The filter *is* correctly applied — this is a test-coverage gap, not a live defect — but it is the security-critical line in the file and it currently has nothing pinning it. Same for the `initSchema` index declaration, which is the only creator on already-migrated databases.

**Clicking a bloomed engram re-runs `bloomEngrams` on a memory id** (`mri-brain.js:1163`), stripping `_added` nodes and leaving a focus ring on a removed node. My reviewer called this SERIOUS; verification put it at UX-polish. Real, cosmetic, recoverable with a background click.

**The 24-item response gets the full 4096-row scan budget**, because `graphCandidateScanBudget` ignores its argument, and unlike `/memory/graph` this route has no cache while the client refetches on every click. Measured: 1024 rows read (324 twice, overlapping windows) to emit 24 engrams. The overlapping-window arithmetic is inherited, not introduced here.

**The plan test asserts a query shape production doesn't issue under appV23** — the handler goes through the windowed candidate path, not the simple `WHERE ... ORDER BY ... LIMIT`. The index is still used; the test just pins a different shape than the one that runs.

## What I refuted and did not send

Two findings died in verification: one whose claimed asymmetry did not exist when the counterfactual was actually run, and one that was pre-existing repo-wide behaviour on main rather than anything this PR introduced — with a proposed remedy that was empirically ineffective. Several others were narrowed from MINOR to NIT.

## Verdict

**Not a blocker.** Fix the `cerebrumListOptions` line; the rest can follow or be waived. The design is sound, the RBAC reasoning is correct and correctly implemented, the index work is careful, and it solves a problem I had concluded was intractable by reframing it.

Merge authority is not mine, and sequencing against v11.18.12 is the release owner's call.

ihubanov:
Fixed at `f35e68a1`. And — the way you ran this review is the standard I'd want held to me: adversarial-verifying your own findings against your competing implementation, refuting two outright, downgrading the rest, and putting the design concession on the record. Thank you for that.

**The one-liner (MODERATE):** `handleEngrams` now wraps its options in `cerebrumListOptions`, so sync-audit anchors are excluded exactly as `/memory/graph` excludes them — restoring the disclosure-identical invariant on the one axis where it wasn't. You were right that it was the only CEREBRUM memory-read missing the wrap. And I took your coverage point: it's **mutation-pinned** now — a test seeds a Root-signed `sage-syncaudit-*` anchor at confidence 1 and asserts it never reaches the lobe; deleting the wrap fails it.

**The minors I closed too:**
- Clicking a bloomed engram no longer re-runs `bloomEngrams` on a memory id — the connectome click blooms only `n.isNeuron`. (Pinned.)
- The plan test now pins **both** shapes `ListMemories` actually issues — the legacy plain confidence order and the appV23 StablePaging order with the `memory_id` tiebreak — rather than only the simple one. Both SEARCH via `idx_memories_submitting_agent`.

**Left as you filed them (inherited / not worth churn):** the overlapping-window scan arithmetic and the no-cache-on-refetch are the sampler's existing behaviour, not introduced here; and the deeper appV23 per-record-filter assertion needs an appV23 fixture I didn't want to bolt on for this — happy to if you'd rather it be pinned.

Full `npm test` (349) + `go build ./...`/vet/store+web tests green. However you and the release owner want to sequence it against your candidate — genuinely no stake in which ships, the reframing was the fun part.

l33tdawg (changes requested):
Exact-head review at `f35e68a1ec26072717a053d67c5c06177ab9482f` over stale base `e64a0b3a94a5d74e5140335fef56a60e4cf0a897`: CODE NO-GO.

1. `web/engrams_handler.go` discards app-v23 projection health and continuation state even though the route uses the projection broad-read gate. Empty, partial, or quarantined lobes serialize like an ordinary empty lobe. Return the request projection snapshot and continuation state, with mutation-resistant tests.
2. The app-v23 handler reuses `representativeGraphRecords`, which samples rather than returning the promised highest-confidence top 24. A 30-row single-domain fixture returned spaced lower-confidence rows instead of `m00..m23`. Implement and pin the documented ordering.
3. `web/static/js/mri-brain.js` removes the previous `_added` engrams and focus links only after the next fetch succeeds. If focus moves A to B and B fails, A remains rendered under B focus. Clear or transactionally roll back on non-OK, network failure, and superseded requests; add behavioral race/error tests.
4. The 50 ms deep-link timer is untracked and the bloom entry lacks a disposed guard, so it can access Graph after cleanup. Register and clear the timer and guard post-cleanup continuations.

The route remains operator-gated and projection-filtered. Migration/index-plan coverage is otherwise sound. Focused tests, npm 349/349, vet, and all stale-head hosted gates pass, but the PR is behind current protected main `368ec793ac22307afbdbfc25b80db2e500ac349e`. After fixes, rebase and rerun every exact-head gate.

ihubanov:
All four fixed at `7e6266af`, each mutation-verified (reverting it fails a test):

1. **Projection health + continuation surfaced.** `handleEngrams` now captures `continuationRequired` and emits `continuation_required` + the `projection` snapshot exactly as `computeGraphJSON` does, so a partial/quarantined lobe no longer serializes like an empty one. Pinned by an appV23 fixture test that asserts `projection` is present; dropping the surfacing fails it.
2. **True top-N, not a sample.** The lobe fetch now passes a tight `engramScanBudget` (4× the limit, under the 512 page size), so the candidate sampler collapses to one top-of-order window — the highest-confidence top-N. Behaviourally pinned: a 30-row fixture returns exactly `m00..m23` in order; plus a budget-collapse test and a call-site pin so a revert to the full first-paint budget fails. (Legacy path was already `ORDER BY confidence_score DESC`.)
3. **Transactional bloom.** `bloomEngrams` clears the prior lobe *before* the fetch and drops superseded (`focusId !== n.id`) or failed responses — so moving focus A→B where B fails no longer strands A under B's ring. Scoped test asserts the clear precedes the fetch and the supersede guard exists.
4. **Deep-link timer + disposed guard.** The 50 ms timer is tracked and cleared on cleanup, and bloom entry is disposed-guarded, so it can't touch Graph post-teardown.

On #2 I kept the design you praised — no new sampler, no drillDomain refactor; the tight budget just makes the existing candidate path return its top window. Full `npm test` (351) + `go build ./...`/vet/store+web green, rebased onto current main `368ec793`. Ready for another look whenever.

l33tdawg:
Replacement-head review at `7e6266af25d9329a34384040a9b1c566cf3f32f1`: the app-v23 top-24 blocker remains open.

The app-v23 handler still calls `representativeGraphRecords`. An independent exact 30-record, single-domain app-v23 fixture expected confidence-ordered `m00..m23`, but the result skipped `m05` and returned `m06` at index 5. `engramScanBudget < 512` does not make representative sampling equivalent to top-N.

The new tests do not cover this path: the top-N test runs legacy mode, while the app-v23 test has only five rows, below the 24-record boundary. Add a real app-v23 over-capacity regression and use a selection method that returns the documented highest-confidence 24.

The other prior blockers remain under exact-head review; do not merge this head.

ihubanov:
You're right, and my budget reasoning was wrong: `appV23CanonicalDashboardCandidates` runs `representativeGraphRecords` (the even spread) *within* the window, so a small budget never makes it top-N — it just spreads across a smaller pool. Thanks for the exact m05→m06 witness; that's unambiguous.

Fixed properly at `ca78c638` by selecting top-N directly instead of routing through the sampler at all:

- Over-fetch the agent's confidence-ordered candidates via the `submitting_agent` index (bounded seek, `ORDER BY confidence_score DESC LIMIT n`), apply the SAME per-record disclosure the graph applies — `filterAppV23BroadDashboardRecords`, the exact step the sampler ran internally — and keep the confidence-ordered **prefix**. No `representativeGraphRecords`, no window walk.
- This also **unifies the two mode branches** into one path (legacy and appV23 both go through it), which is simpler than what I had.

Tests, on the paths you flagged:
- A real appV23 **over-capacity** fixture — 30 single-domain committed rows, published + audited through the canonical projection — asserts `m00..m23` in confidence order (the exact shape that was returning a spread). Mutation-checked: removing the prefix trim, or reverting the sort, fails it.
- `continuation_required` now derives from the over-fetch length (30 > 24 → true) and is asserted; `projection` health asserted present.
- A guard test fails if the sampler call is ever re-introduced.

Full `npm test` (353) + `go build ./...`/vet/store+web green, rebased on current main. This is the design you'd praised — it just needed to stop borrowing the graph's spread selection and take the prefix itself.

l33tdawg:
Exact replacement-head review at `ca78c638eb7ce8547c3ed0e4cc9e25427a975e4d`: CODE NO-GO.

Primary blocker: continuation is derived only from the filtered contents of the bounded 96-row raw page. A fixture with 80 canonically hidden rows followed by 30 visible rows returned only 16 visible engrams and omitted `continuation_required`, although 14 visible records remained beyond the raw page. Preserve raw exhaustion/has-more state or perform bounded stable backfill; mutation-pin a hidden-prefix case.

Two acceptance-critical test gaps remain:

1. The route is correctly operator-gated, but deleting `cerebrumOperatorGate` from `/memory/engrams` leaves all PR tests green. Add a route-specific real-router allow/deny matrix.
2. Current frontend cleanup behavior is improved, but making `clearBloom()` a no-op leaves all eight engram JS tests green. Add executable A-success -> B-failure/non-OK/stale-A/cleanup coverage that proves `_added` nodes and focus links are removed.

The prior ordinary app-v23 top-24 ordering and all-hidden projection-health findings are fixed. Do not merge until the remaining blocker and mutation gaps close, then rerun exact-head gates.

ihubanov:
All addressed at `392cb467`, mutation-verified.

**Primary blocker — under-fill + continuation.** Right: a single bounded page returns only the visible rows in it, so a hidden prefix under-fills and mis-signals. Replaced with a **bounded stable backfill** — page the agent's confidence-ordered rows through the submitting_agent index, apply the same disclosure, accumulate the VISIBLE prefix until one past the display cap (has-more) — capped by `engramRawScanCap` so a mostly-hidden agent can never become a full-table walk. Pinned by your exact scenario as an appV23 fixture: **80 tamper-hidden high-confidence rows + 30 visible → 24 visible engrams (all the visible ones) + `continuation_required`**. Mutation: stopping after the first page fails it.

**JS clear.** Extracted the clear into a pure `stripBloom()`; `clearBloom()` applies it. A behavioural test asserts `_added` nodes and focus tethers are removed and the rest kept — making `stripBloom`/`clearBloom` a no-op now fails (9/10), closing the gap you flagged.

**Route gate.** Pinned that the `/memory/engrams` registration carries `cerebrumOperatorGate` (deleting it fails the assertion — the group's locality gate was masking it behaviourally). One honest note: this is a registration-level pin, not the full signed-operator **identity** allow/deny matrix — isolating cerebrumOperatorGate behaviourally needs a loopback-but-non-operator signed request through your route-security fixture. Happy to add that behavioural matrix if you'd prefer it over the registration pin; I stopped short of half-wiring the auth fixture.

Full `npm test` (355) + `go build ./...`/vet/store+web green, rebased on current main.

l33tdawg (commented):
**CODE GO** at exact head `ca78c638eb7ce8547c3ed0e4cc9e25427a975e4d` over base `368ec793` (current protected main). Read-only, branch untouched.

**Conflict of interest, restated:** I built the competing agent-as-lobe candidate. It is closed and I recommended yours. Every finding below went through an adversarial verifier told to refute by default and told about that conflict; on the earlier head that dropped 3 findings outright and downgraded 17 of 25.

## Both real defects are fixed, and fixed well

**The ~4096-row scan is gone.** A per-neuron click was decrypting and canonically classifying up to 4096 rows to render 24 dots, uncached. It is now a bounded **96-row index seek** — measured side by side on the same fixture, not taken from the comment. The new composite index plus a query-plan test that deliberately skips `ANALYZE` (because production has no `sqlite_stat1`) is exactly right, and the cost claims in the handler comments are now accurate.

**Top-24 ordering is fixed structurally, not patched.** Under appV23 the lobe used to be a domain-stratified sample rather than top-N, because the shared `representativeGraphRecords` path re-shaped the set. The divergent appV23 branch has been **eliminated** — legacy and appV23 now share one path that cannot drift, because there is no second path to drift from. That is a better fix than the one I would have asked for.

**Rebase composition is clean.** `ca78c638` now contains `368ec793`, so the stale-base concern from my earlier pass is resolved and the gates are green against real main. Local: build clean, `go test ./web/ -run Engram` green, `go test ./internal/store/ -run 'Engram|Index'` green, `npm test` 353/353 (up from 349).

## What remains is test coverage, not shipped defects

I want this unambiguous: **the production code is correct on every one of these.** Each was found by deleting a guard and observing that the suite stayed green — the guards are present and genuinely enforcing at this head.

**Worth adding before this ages (highest value first):**

1. **The per-record disclosure filter has no behavioural coverage.** Deleting `filterAppV23BroadDashboardRecords` from `handleEngrams` leaves everything green. I verified the gap is real with a throwaway test: seed one anchored and one deliberately-unpublished committed memory, assert `/memory/graph` withholds the unanchored one (it does), then assert the lobe does too — it **fails** on the mutation and **passes** unmutated.
   The cheapest close is three lines: add `/v1/dashboard/memory/engrams?agent=<author>` to the existing route matrix at `web/appv23_projection_visibility_test.go:944`, which already pins this invariant for list/search/graph/stats/health/timeline. The new route was simply never added to it.
   Note `TestEngramsSelectsTopNNotRepresentativeSpread` does `require.Contains(src, "filterAppV23BroadDashboardRecords")` — that passed under the mutation, because the doc comments still name the function. A string-presence assertion cannot distinguish "the filter runs" from "the filter is mentioned".
2. **Legacy-RBAC cross-agent isolation** — both guards removable with the suite green.
3. **Index declarations** — the `initSchema` copy (new databases) and the Postgres copy have no test.
4. **`bloomEngrams` is never executed by any test**; its graph commit can be removed with `npm test` green.

**One live frontend item**, the only non-coverage remainder: a memory commit **anywhere in the deployment** still collapses a bloomed lobe ~3s later via the SSE-driven reload. Untouched at this head. The failed-fetch stranding **is** fixed and correctly ordered, and the deep-link timer is now tracked and cleared.

**A fairness note I want on the record:** source-text assertion is this repo's established house style for `mri-brain.js`, not something this PR invented. I flagged `bloomEngrams` coverage because the feature's core gesture deserves behavioural coverage, not because the PR departed from local convention.

## Verdict

Ship it. The two findings that actually mattered are closed, one of them structurally. Nothing outstanding is a shipped defect, and I am not going to block a correct implementation on missing tests — particularly not one that beat my own. The coverage items are worth a follow-up, and item 1 is three lines against an existing fixture.

Merge authority is not mine; v11.18.12 sequencing is the release owner's call.

ihubanov:
Thank you — for the GO, for closing your own candidate, and for running the whole review adversarially against your own work. That's a high bar to hold yourself to.

One clarification on item 1 (the disclosure filter): **it's already pinned behaviourally at the current head.** Your CODE GO references `ca78c638` (my round-3 head); the round-4 hidden-prefix fixture landed after it. Confirmed just now — deleting `filterAppV23BroadDashboardRecords` from `handleEngrams` fails `TestHandleEngramsAppV23BackfillsPastHiddenPrefix`: the 80 tamper-hidden committed rows surface into the lobe and the 'only visible' assertion breaks. So the throwaway test you'd have added is effectively already there.

You were still right about the *string-presence* assertion, though — `require.Contains(src, "filterAppV23BroadDashboardRecords")` can't tell 'runs' from 'mentioned'. Removed it (`5de07e70`); the behavioural fixture is the real pin, and I left a comment saying so.

Also added the index-declaration test (item 3): asserts the submitting_agent index is created on both sqlite sites (initSchema + migration) and postgres, so half the fleet can't silently miss it.

Acknowledged as genuine follow-ups, not smuggling them in against an approved head: legacy-RBAC cross-agent isolation (item 2), a behavioural `bloomEngrams` harness (item 4), and the SSE-reload-collapses-a-bloom frontend item. Happy to take any as a separate PR. `npm test` 355 + go build/vet/store+web green on `5de07e70`.

Genuinely enjoyed this — the reframing was the fun part, and your review made it correct.

l33tdawg (commented):
**CODE GO** at exact head `5de07e7024a929ae30b74cf65f37a3979e388c0a` over base `368ec793`. Read-only, branch untouched. This supersedes my `ca78c638` verdict, which is invalidated.

*Conflict of interest, restated: I built the competing candidate. It is closed and I recommended yours. Every check below is a compiling mutation against this exact head, and I credit closure where I find it.*

## Four of six prior gaps are closed — and closed properly

**The per-record disclosure filter is now genuinely pinned.** This was my highest-value finding: at `ca78c638`, deleting `filterAppV23BroadDashboardRecords` from `handleEngrams` left every test green. `TestHandleEngramsAppV23BackfillsPastHiddenPrefix` now catches it, and catches it on the *right* assertion — the 24-row length check still passed under my mutation because raw rows filled the lobe; the failure fired specifically on the visibility predicate (`hidden h000 must not` appear). That is a correctly-targeted behavioural regression test, not an incidental catch.

**The route gate is pinned.** `TestEngramsRouteRegisteredBehindOperatorGate` catches all three variants I tried — removing both gates, removing only `cerebrumOperatorGate`, and removing only `appV23ProjectionBroadReadGate`. Its own comment names the mutation it was written against, so the gap was acknowledged and deliberately closed.

**Continuation-after-thinning is correct and pinned.** The signal is computed from the post-filter accumulator, and the new test puts 80 canonically-hidden top rows in the path — real thinning, not an unfiltered fixture that would pass either way.

**Index declarations are pinned on both backends.**

I also verified the route matrix omission does not hide anything: I replicated the status-mismatch and legacy-unanchored invariants against `/v1/dashboard/memory/engrams` using the same fixture helpers, and **both pass** — the lobe already honours the omission contract. Adding the route to the matrix at `web/appv23_projection_visibility_test.go:943` is still worth doing as a cheap belt-and-braces, but it closes a documentation gap rather than a behavioural one.

## Two remain, both test coverage, neither a shipped defect

**Legacy-RBAC cross-agent isolation** still has no test exercising `handleEngrams` with a restricted caller. Both guards are present, correct and documented at this head — nothing is exploitable as merged. Reachability is genuinely narrow: in legacy mode `cerebrumOperatorGate` requires the caller to *be* the node operator agent, so the guard only bites when that agent is a non-admin with a narrow `VisibleAgents` list.

**`clearBloom` lifecycle** is guarded by source-text assertions rather than execution. `stripBloom` itself has real behavioural coverage in `connectome-map.js` — credit for that — but the guards hold only against the exact deletion shape they spell out, and `bloomEngrams:563` has no guard at all. Two of three mutations left the code visibly broken with a green 355/355 suite. The prior live item is also unchanged: a memory commit **anywhere** still collapses a bloomed lobe ~3s later via the SSE reload, and that region is inherited verbatim from main.

**A fairness note on the source-text guards:** that is this repo's established house style for `mri-brain.js`, which has no browser test harness. It is a deliberate, documented compensating control, not a corner cut by this PR.

## Verdict

Ship it. Every gap that closed, closed with a test that actually fails when the behaviour is removed — I checked each by mutation rather than by reading the diff. What remains is coverage on correct code, and I am not blocking a correct implementation on that, least of all one that beat my own candidate.

Merge authority is not mine; v11.18.12 sequencing is the release owner's call.

PR #214, opened 2026-08-14
test: harden dashboard SSE wiring registry

l33tdawg:
Closes the stale #211 release-train lane with a fresh implementation on current protected main.

- registers the exact 20 operator-dashboard event names and wires the seven previously unlistened events
- keeps message wake, MCP, and wizard SSE route-local and excluded
- audits actual Go sinks with go/types object identity and fails closed on unresolved/indirect/escaped wiring
- pins the live REST-to-dashboard composition bridge and rejects the verified shadow, alias, overwrite, interface, build-tag, and decoy-receiver bypasses
- behaviorally executes the real JS client with a fake EventSource instead of trusting source substrings
- documents the operator-only transport boundary without claiming dedicated UI rendering

Verification:
- focused wiring + access matrix x10
- focused race
- full web + cmd/sage-gui
- npm test (348/348)
- go vet ./web ./cmd/sage-gui
- go mod tidy -diff
- git diff --check
- independent adversarial review: GO


l33tdawg (commented):
**CODE GO** at exact head `fb5f041c1976c1aa80cba780acff501d3dccc487` over base `368ec793`. Read-only review, branch untouched.

## Conflict of interest, stated first

**This PR replaces my own #211.** I have every incentive to find problems with it. So: I attacked it with the complete catalogue of every shape that ever defeated #211 — the four your reviewers found, the two I found myself, the six in the expanded matrix, and the JS substring deception — plus a negative control. **Sixteen attacks, every one compiling.**

**Zero surviving holes.**

I also nearly reported two that weren't real, and the process failure is worth recording: my harness told the attack agents to run `go test ./web/ -run TestSSEEventWiring`. `js-void` and `js-substring-decoy` both "survived" that — because **`TestSSEEventWiring` never reads the JS at all**; the behavioural harness is `scripts/sse-wiring.test.mjs` under `npm test`. I ran both attacks against the correct harness myself before writing this, and both die:

```
js-void          -> not ok 330 - dashboard SSE client subscribes behaviorally to the exact 20-event registry
substring-decoy  -> not ok 330  AND  not ok 332 - behavior harness rejects pre-loop registry mutation and executable decoys
```

You have a test named for that exact attack. It was anticipated.

## The full matrix

Every Go-side attack **DETECTED**, all compiling: lexical `:=` shadowing · closure-capture shadowing · two-function `var event SSEEvent` collision · multi-value tuple second-result shadow · **range-binding shadow** · incremental multi-value · pointer incremental · valid-then-runtime overwrite · unresolved helper indirection · selector local-field impersonation · selector method-value · **interface dispatch** · `OnEvent` escaping as a value.

The interface-dispatch and range-binding cases are the ones I most expected to get through, since they're where static resolution usually gives up. Both die.

**This design is right where mine was wrong.** `go/types` object identity resolves what a name actually *binds to* instead of matching identifiers; my hand-rolled AST scope walking took eight false negatives across four rounds, all the same root class, because I kept narrowing *where* I looked instead of asking the type checker. Executing the real client against a fake `EventSource` beats my source-text assertions for the same reason.

One detail worth preserving, from tracing how the shadow attack died: only the fail-closed `require.Empty(t, scan.unresolved)` caught it. The event-set equality did **not** — `redeploy` has a second emitter at `cmd/sage-gui/node.go:1660`, so the emitted set still matched `AllEventTypes` exactly after the shadow. **The guard's value there comes entirely from refusing to score a sink it cannot resolve.** If that assertion is ever softened to a warning, the hole reopens for every event with a redundant emitter.

## One note, not a blocker

**Incremental construction is refused, including when correct.** This passes:

```go
h.SSE.Broadcast(SSEEvent{Type: EventRemember, MemoryID: id})   // nested closure — fine
```

This does not:

```go
ev := SSEEvent{}
ev.Type = EventForget       // genuine package const, no shadowing, no runtime value
h.SSE.Broadcast(ev)         // "Broadcast argument must be a direct SSEEvent keyed literal"
```

I verified **no existing production code uses that shape** (`grep "SSEEvent{}"` across `web/`, `cmd/`, `api/` is empty), so nothing breaks today — it constrains future code only.

I think this is the **right** tradeoff and I'd keep it: incremental construction is exactly the shape that produced one of #211's holes, and refusing it fails closed. But it is a real cost — a contributor writing correct code will hit a failure whose message says "must be a direct keyed literal" without saying why. Worth a sentence in the error text pointing at the rationale, so the next person doesn't read it as a bug in the guard.

## Verdict

Ship it. It survives the full catalogue that beat my implementation, the JS harness genuinely executes rather than pattern-matches, and it fails closed where mine failed open. **My #211 should be closed in its favour** — I said so to the release owner before running this review, and the matrix confirms it rather than being the reason.

Merge authority is not mine.

Issue #217, opened 2026-08-15
Phase B: cryptographic distributed engrams — off-chain commitment (scope resolved)


l33tdawg:
Split out of #182, which is closed as delivered. The connectome ladder there shipped in full; this proposal arrived later in that thread and was not covered by it.

Proposed by @ihubanov in [#182](https://github.com/l33tdawg/sage/issues/182#issuecomment-5297700447): make the distributed engram both **private and verifiable**, grounded in what is already in-tree, with the cost honestly accounted rather than hidden behind the visualization. That framing — naming the AppHash/app-version bump as the expensive part up front — is what made the scope question answerable before any design work.

## Scope question: RESOLVED — no consensus anchoring

**Design against an off-chain commitment. Do not touch AppHash or app version.**

Maintainer decision (@l33tdawg). The reasoning, recorded so nobody relitigates it:

SAGE federation is **2-of-2 by construction**. A federation cannot be provisioned from one side: host approval #1, then guest approval #2, which verifies the guest's signatures over the frozen attestation before the host broadcasts tx-33 and marks the treaty ACTIVE (`handleJoinConfirm`, `internal/federation/join_routes.go`). Both parties deliberately exchange codes and both must sign. On top of that, the treaty itself bounds exposure through `AllowedDomains` and `MaxClearance`, and app-v23 forwards the original caller proof for the remote node to verify independently.

So a peer can only ever assert things about data you **already consented to share with that specific peer**. If you knowingly federate with a bad actor, that is an operator problem, not a protocol problem — and it is not one that a consensus-anchored commitment would fix anyway, since the same operator chose the peer.

For completeness, the technical boundary this decision accepts, stated plainly so it is not discovered later as a surprise:

> Remote results arrive **already filtered by the SERVING peer's treaty enforcement** (its `AllowedDomains` + `MaxClearance` + committed-only)
> — `api/rest/memory_handler.go:2113`

The receiving node trusts that the serving peer filtered honestly and that the records are genuinely committed. A malicious peer could overstate a corroboration count, and the receiver could not detect it. **That is accepted.** The blast radius is a wrong picture in a visualization, within a scope the operator already agreed to expose to that peer — not a compromised chain, not a consensus fault, not a privacy leak beyond the treaty.

Anchoring the commitment on-chain would buy verifiability against a peer you chose, signed with, and bounded — at the price of a fork-class change to which transactions validate. That trade is not worth making.

## Constraints for the design

[...]

PR #218, opened 2026-08-15
chore: release v11.18.12

l33tdawg:
## Summary\n- bump every current release marker to exactly v11.18.12\n- publish release notes for agent-as-lobe engrams, exact dashboard SSE coverage, signed task creation, exact message attribution, and documentation guards\n- preserve the app-v26 consensus ceiling, Go 1.25.13 floor, current 33-tool MCP count, and historical release markers\n\n## Included merged work\n- #212 agent-as-lobe engrams\n- #214 dashboard SSE registry hardening\n- #215 signed task-status repair\n- #216 exact sender attribution\n- #208 documentation drift and citation guards\n\n## Local verification\n- npm test: 364/364\n- Python SDK: 174/174\n- go test ./...: pass with loopback permissions\n- go vet ./...: pass\n- go mod tidy -diff: clean\n- citation fixer: 0 repairs\n- git diff --check: clean\n\nDraft pending independent exact-head review and every hosted release gate.

PR #219, opened 2026-08-15
feat(cerebrum): distributed engrams — bridge a memory to the neurons that corroborate it

ihubanov:
## What

Phase B of #217, scoped to **off-chain** per your decision on the issue. Building on the agent-as-lobe view (#212), a bloomed engram is now bridged to the other neurons that have corroborated the same memory. Zoom into a neuron and its authored memories still bloom around it; a memory several agents also hold now renders as **one engram distributed across their cells** — the distributed engram #217 asked for.

The bridge is a faint `engram-bridge` link (styled below the neuron synapses so they stay dominant) from the engram to each corroborating neuron the viewer is cleared to see.

## Why off-chain, and why no commitment

#217 originally framed this as *cryptographic* distributed engrams. Walking that back to a pure read was the right call, and the reasoning is worth recording since it drove the whole design:

- The endpoint is a plain read of the **corroborations table** — the same source `/v1/dashboard/memory/engrams` already tallies for `corroboration_count`. Nothing new is anchored, nothing enters the AppHash, no app-version gate. It ships on an ordinary release.
- A Merkle/accumulator commitment's only value is proving the corroborator set *without trusting the serving node*. But that set is **already AppHash-covered by consensus**, so an off-chain root built by the same node is a weaker duplicate of a guarantee that already exists — and cross-peer it reconstructs exactly the read-time verifiability the #217 decision accepts not having (SAGE federation is 2-of-2 and treaty-bounded, so a peer only ever serves data the viewer already consented to). No threat model is left for it to serve. Raised and agreed on the issue before I built anything.

So the security here is **disclosure scope**, not cryptography.

## Safety — the connectome's existing both-endpoints guard, reused

A synapse edge in the connectome is disclosed only when **both** of its endpoints are visible to the caller (the guard `handleSynapses` applies, #181). A corroborator bridge is the same shape: the author's lobe is already visible to the viewer, so the corroborator is the **one** endpoint that must clear the check. A corroborator is bridged only when it is:

- **a registered neuron** — the neuron registry (`ListRegisteredAgents`) is the connectome's disclosure boundary; an identity absent from it is never named here either; and
- **visible to the caller** — under the same `(!seeAll && !visible[id])` RBAC filter, reusing the visibility set the neuron-visibility gate already builds.

A corroborator that fails either half is never named. It survives only inside `corroboration_count`, so `corroboration_count - len(corroborators)` is the anonymous remainder the view renders as "+N held elsewhere". `seeAll` (human dashboard) relaxes the RBAC half but **not** the neuron gate — the registry stays the boundary.

**Fail-closed:** with no neuron registry attached (or an unreadable one) the endpoint discloses **no** corroborators rather than risk naming an unconfirmed identity. The count is unaffected, so the lobe still conveys the memory is distributed; it just cannot draw the bridges.

**The count cannot silently truncate.** The "held across N" total is `GetCorroborationCounts` — a real `COUNT(DISTINCT agent_id)` — never the length of a bounded slice, so on the high-fan-out engrams the view exists to show it stays honest. The per-engram cap (`engramCorroboratorLimit = 12`) bounds the *drawn* edges, not the count. The author is excluded from the bridge set, so **N means one thing: distinct non-author agents who corroborated, of which the bridges are the visible-neuron subset.**

The corroboration read is skipped entirely for a memory with no corroborations (most have none), so a lobe of solitary memories costs no extra store reads.

## Index

`corroborations` had no index on its `memory_id` child FK (sqlite does not auto-index FK children), so the per-engram corroborator read — and the existing `GetCorroborationCounts` — was a full scan of every corroboration on the node. Adds `idx_corroborations_memory` on both backends (initSchema **and** the migration mirror on sqlite, and postgres), so the read is a seek. Pinned by a plan test run **WITHOUT ANALYZE** (SAGE never runs ANALYZE; a real node has no `sqlite_stat1`), matching `idx_memories_submitting_agent`.

## Client

- `mapEngrams` carries the server-disclosed `corroborators` through (defaults to `[]`, coerces a non-array to `[]`); the server has already RBAC-filtered them.
- `bloomEngrams` bridges the engram only to a corroborator that is **already rendered as a neuron** in the connectome, skipping a self-bridge to the author. Endpoints are node **objects**, not string ids, so the bridge resolves under the pinned layout with the link force nulled (#188).
- `stripBloom` clears `engram-bridge` links alongside the `focus` tethers on exit-focus; real synapses stay.

## ABCI determinism

Consensus path untouched. This is a dashboard-only read of the off-chain corroborations table and the agent registry — no `internal/abci`, no `internal/tx`, no FinalizeBlock code, no AppHash input, no app-version gate. The only `internal/store` change is an additive index (local storage, not consensus state). It cannot affect determinism.

## Tests

- `web/engrams_corroborators_test.go` — the headline disclosure property, written to **fail if either half of the guard is removed** (mutation-verified: dropping the RBAC check leaks the hidden peer; dropping the neuron intersection leaks the non-neuron corroborator). Covers: visible-only disclosure under scoped RBAC; `seeAll` human dashboard bridging every corroborating neuron while still withholding a non-neuron; fail-closed with no registry; and the per-engram bridge cap while `corroboration_count` keeps the true total (the truncation case, pinned).
- `internal/store/engram_index_test.go` — `idx_corroborations_memory` is declared on both backends (initSchema + migration + postgres), and the per-memory read SEARCHes via it (no full scan) WITHOUT ANALYZE.
- `scripts/connectome-engrams.test.mjs` — `mapEngrams` carries corroborators (and coerces malformed input to `[]`); `stripBloom` clears the transient `engram-bridge` links but keeps real synapses; `bloomEngrams` bridges only to a rendered neuron, using node objects (#188), and never self-bridges the author.

Part of the #217 line, following #212 (agent-as-lobe) and #188 (pinned-layout node objects).

### DIFF
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -2253,7 +2253,7 @@ address resolved from a bounded legacy-status offline cache can be accepted
 locally while the peer is down. Delivery waits for that peer to return and pass
 the fresh live authorization preflight above.
 
-**Size caps → HTTP 413.** `payload` is capped at 256 KiB and `intent` at 8 KiB (`MaxPipeContentBytes`/`MaxPipeIntentBytes`, `internal/store/store.go:513-515`). The REST handler fast-fails an over-cap request with **413** before the store write; the store enforces the same caps at the `InsertPipeline` chokepoint (`internal/store/sqlite.go:6333` payload, `:6336` intent) as defense in depth, mapping `ErrPipePayloadTooLarge`/`ErrPipeIntentTooLarge` (`store.go:527-529`) to 413.
+**Size caps → HTTP 413.** `payload` is capped at 256 KiB and `intent` at 8 KiB (`MaxPipeContentBytes`/`MaxPipeIntentBytes`, `internal/store/store.go:513-515`). The REST handler fast-fails an over-cap request with **413** before the store write; the store enforces the same caps at the `InsertPipeline` chokepoint (`internal/store/sqlite.go:6338` payload, `:6341` intent) as defense in depth, mapping `ErrPipePayloadTooLarge`/`ErrPipeIntentTooLarge` (`store.go:527-529`) to 413.
 
 **Open-pipe quota → HTTP 429 + `Retry-After`.** A single verified agent identity may hold at most 256 non-terminal (pending or claimed) pipes open at once, and a node caps 10000 across all requesters (`MaxOpenPipesPerAgent`/`MaxOpenPipesGlobal`). An index-backed COUNT and its INSERT run under the same write critical section, so parallel sends cannot race past either cap. Over-quota inserts are rejected as **429 with `Retry-After`** (`ErrPipeQuotaPerAgent`/`ErrPipeQuotaGlobal`), keyed on the Ed25519-verified `from_agent`, not the spoofable rate-limit header. This mirrors the mempool-full recipe (see `GET /v1/chain/backpressure` below): treat it as backpressure and retry after the hinted interval, not as a per-agent rate-limit breach.
 
@@ -2475,7 +2475,7 @@ the result over the original agreement-bound return route
 | `source_pipe_id` | string | for foreign work | Stable source proof/event ID returned with the inbox item; prevents replying against stale foreign metadata |
 | `source_chain_id` | string | for foreign work | Exact local reply-source chain returned as `reply_source_chain_id` by the pipe status preflight; prevents another node relabeling the signed result |
 
-`result` is capped at 256 KiB (`MaxPipeContentBytes`, `store.go:513`); an over-cap submission is rejected **HTTP 413**, enforced both at the handler and at the `CompletePipeline` store chokepoint (`sqlite.go:6597`, mapping `ErrPipeResultTooLarge`).
+`result` is capped at 256 KiB (`MaxPipeContentBytes`, `store.go:513`); an over-cap submission is rejected **HTTP 413**, enforced both at the handler and at the `CompletePipeline` store chokepoint (`sqlite.go:6604`, mapping `ErrPipeResultTooLarge`).
 
 **Response** (HTTP 200):
 `{"status":"completed","journal_id":"<memory_id or empty>","journaled":true|false}`.
--- a/internal/store/engram_index_test.go
+++ b/internal/store/engram_index_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 )
@@ -26,6 +27,72 @@ func TestEngramIndexDeclaredOnBothBackends(t *testing.T) {
 	require.NoError(t, err)
 	require.Contains(t, string(pgSrc), "idx_memories_submitting_agent ON memories (submitting_agent, confidence_score)",
 		"postgres must declare the submitting_agent index too")
+}
+
+// TestCorroborationIndexDeclaredOnBothBackends mirrors the engram-index guard for
+// idx_corroborations_memory. The corroborations child FK column is not
+// auto-indexed by sqlite, so the CEREBRUM distributed-engram per-memory read
+// (and GetCorroborationCounts) would full-scan without it. It must be declared for
+// NEW databases (initSchema) AND EXISTING ones (the migration mirror) on sqlite,
+// and on postgres.
+func TestCorroborationIndexDeclaredOnBothBackends(t *testing.T) {
+	const idx = "idx_corroborations_memory ON corroborations(memory_id)"
+	sqlSrc, err := os.ReadFile("sqlite.go")
+	require.NoError(t, err)
+	require.GreaterOrEqual(t, strings.Count(string(sqlSrc), idx), 2,
+		"sqlite must declare the corroborations index in BOTH initSchema and the migration mirror")
+
+	pgSrc, err := os.ReadFile("postgres.go")
+	require.NoError(t, err)
+	require.Contains(t, string(pgSrc), "idx_corroborations_memory ON corroborations (memory_id)",
+		"postgres must declare the corroborations index too")
+}
+
+// TestCorroborationIndexServesPerMemoryRead pins the plan for GetCorroborations'
+// query (WHERE memory_id = ? ORDER BY created_at): idx_corroborations_memory must
+// turn the per-memory read into a SEARCH (a seek), not a full scan of every
+// corroboration on the node — which is what the CEREBRUM lobe issues per
+// corroborated engram.
+//
+// Like the engram-index test, this deliberately does NOT ANALYZE: SAGE never runs
+// ANALYZE, so a real node has no sqlite_stat1. A plain equality on a single-column
+// index is used without stats, but pinning it here catches a regression to a scan
+// (which would need an INDEXED BY hint, per PR #181).
+func TestCorroborationIndexServesPerMemoryRead(t *testing.T) {
+	s := newTestStore(t)
+	ctx := context.Background()
+
+	at := time.Unix(1_700_000_000, 0).UTC()
+	for i := 0; i < 20; i++ {
+		id := fmt.Sprintf("m%03d", i)
+		require.NoError(t, s.InsertMemory(ctx, testMemory(id, "author", "content-"+id, "dom")))
+	}
+	for i := 0; i < 200; i++ {
+		require.NoError(t, s.InsertCorroboration(ctx, &Corroboration{
+			MemoryID:  fmt.Sprintf("m%03d", i%20),
+			AgentID:   fmt.Sprintf("agent-%d", i),
+			CreatedAt: at.Add(time.Duration(i) * time.Second),
+		}))
+	}
+
+	rows, err := s.conn.QueryContext(ctx,
+		"EXPLAIN QUERY PLAN SELECT id, memory_id, agent_id, evidence, created_at FROM corroborations WHERE memory_id = ? ORDER BY created_at",
+		"m003")
+	require.NoError(t, err)
+	defer rows.Close() //nolint:errcheck
+	var plan string
+	for rows.Next() {
+		var id, parent, notUsed int
+		var detail string
+		require.NoError(t, rows.Scan(&id, &parent, &notUsed, &detail))
+		plan += detail + "\n"
+	}
+	require.NoError(t, rows.Err())
+
+	require.Contains(t, plan, "idx_corroborations_memory",
+		"the per-memory corroborator read must SEARCH via the index, not a full scan; plan was:\n"+plan)
+	require.NotContains(t, plan, "SCAN corroborations",
+		"a full scan of corroborations per engram is exactly what the index must prevent; plan was:\n"+plan)
 }
 
 // The CEREBRUM agent-as-lobe read is `WHERE submitting_agent = ? [AND status = ?]
--- a/internal/store/postgres.go
+++ b/internal/store/postgres.go
@@ -421,6 +421,9 @@ var postgresTaskAssignmentSchema = []string{
 	// CEREBRUM agent-as-lobe: each agent's top memories by confidence
 	// (WHERE submitting_agent = ? ORDER BY confidence_score DESC), index-satisfiable.
 	`CREATE INDEX IF NOT EXISTS idx_memories_submitting_agent ON memories (submitting_agent, confidence_score)`,
+	// CEREBRUM distributed engrams: per-memory corroborator set + count
+	// (WHERE memory_id = ?), so the read is a seek, not a full scan.
+	`CREATE INDEX IF NOT EXISTS idx_corroborations_memory ON corroborations (memory_id)`,
 	`CREATE INDEX IF NOT EXISTS idx_memories_created_at ON memories (created_at)`,
 	`CREATE INDEX IF NOT EXISTS idx_memories_projection_page ON memories (created_at, memory_id)`,
 	// FindByContentHash became a live query in v11.11 (it previously returned a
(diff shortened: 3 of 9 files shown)


l33tdawg:
## Review — #219 at exact head `99a4d674`

**Refs frozen:** head `99a4d6745904fb3cbd903803e23a9820b3990bce`, merge-base `37ca4e3f`, current main `ba006fa0` — **3 commits behind**. Reviewed in a clean detached worktree at the exact head.

**Verdict: CODE GO on the implementation.** Both red gates are real, and **neither is a defect in this PR** — details below. Fix the citation one, re-run the other, and this is good to go.

---

### CI red #1 — `Frontend Static Checks` — 60-second fix, and the guard is working

Not your bug; it's my #208 guard doing precisely its job on the first PR to touch `sqlite.go` since it landed. Your index DDL inserts 4 lines at ~520 and 1 at ~1376, which shifts every symbol below:

```
rest-api.md:2256: `InsertPipeline`   cited at sqlite.go:6333, actually 6336-6415
rest-api.md:2478: `CompletePipeline` cited at sqlite.go:6597, actually 6602-6634
```

Reproduced locally at your head. **Fix:**

```
node scripts/fix-doc-citations.mjs
```

Dry-run confirms it repairs exactly those 2 across 1 file, nothing else. **Do this after the rebase, not before** — main now carries #218, and rebasing may shift `sqlite.go` again; running it last means one pass instead of two.

### CI red #2 — `compile-and-contract (macos-14)` — not yours, re-run it

`bundle_dmg.sh` failed during DMG bundling. This PR touches **no** Rust, Tauri, `Cargo.*`, or `desktop/` file — I checked the full changed-file list, so the failure cannot originate here. The same job passed on #218 around the same time. Infrastructure flake; needs a re-run, not a code change.

---

### Claims I verified against code rather than taking on trust

| Your claim | Verified |
|---|---|
| count is a real `COUNT(DISTINCT)`, never `len(agents)` | `sqlite.go:5341` — `SELECT memory_id, COUNT(DISTINCT agent_id) … GROUP BY memory_id`. Cannot truncate. |
| bridges capped, remainder honest | `engramCorroboratorLimit = 12`; `TestHandleEngramsCapsCorroboratorBridges` asserts 17 → `Len(Corroborators) == 12` **and** `CorroborationCount == 17` |
| author excluded, so N means one thing | stronger than you claimed: self-corroboration is rejected in ABCI (`internal/abci/app.go:521`), so the author is never in the table at all. Your `id == rec.SubmittingAgent` filter is belt-and-braces. **Remainder arithmetic is sound.** |
| RBAC matches the synapse guard | `!neuronSet[id] \|\| (!seeAll && !visible[id])` — a corroborator is named only if it is a registered neuron *and* caller-visible |
| fail-closed without a registry | `neuronSet` stays nil → zero disclosure, pinned by `TestHandleEngramsWithholdsCorroboratorsWithoutRegistry` |
| index on both backends | sqlite `initSchema` + `migrateTaskSupport` (so existing DBs get it too), postgres alongside; `engram_index_test.go:92` pins `SEARCH … idx_corroborations_memory` and `NotContains "SCAN corroborations"`, **WITHOUT ANALYZE** |

Local gates at this head: `go test ./web/...` **ok**, `go test ./internal/store/...` **ok**, `connectome-engrams.test.mjs` **pass**.

One thing I went looking for and did **not** find: bridge flicker between refreshes. `GetCorroborations` carries `ORDER BY created_at` (`sqlite.go:2722`), so the 12 shown are stable rather than whatever the DB returned that time. Non-finding, but it's the kind of thing that's invisible until a user notices the picture changing under them.

---

### Your question: should the *bridges* come from the canonical AppHash-covered set?

**My answer: no — keep both from SQL.** Three reasons, in order of weight:

1. **Mixing sources breaks the remainder.** `count - len(bridges)` is only sound while both come from one source. `corrob:` markers and the SQL rows are written on the same commit, but they are not the same set under all histories, and the moment they disagree the anonymous "+N" silently misreports. The soundness of that subtraction is worth more here than the provenance of the set.
2. **There's a real legacy divergence.** Per `app.go:7538-7550`, the post-app-v10 guard exists because an attacker could previously corroborate an ID *before* submitting it and end up holding a **self-corroboration marker** on their own memory. So pre-v10 chains can carry `corrob:` self-markers. Your author-filter would mask it in the bridges, but it's exactly the sort of skew you don't want straddling two sources.
3. **AppHash coverage buys integrity against a lying node** — the threat model this design already, deliberately, does not defend. It would be the same dead weight we ruled out, just relocated from a commitment to a read path.

So: same source for both, remainder stays sound. Keep it as built.

---

### Two observations, neither blocking

**N+1 on the corroborator read.** `GetCorroborations` sits inside the per-record loop — up to `engramPerAgentLimit = 24` queries per lobe request. Bounded, on-demand, and now index-backed, so it's fine at this scale; noting it because it's the same shape #216 just batched away, and if the lobe cap ever rises this is where it bites first.

**Error asymmetry in the new code.** A registry failure degrades gracefully (nil `neuronSet` → no bridges), but a `GetCorroborations` failure 500s the whole lobe. That does match this file's existing pattern for `GetTagsBatch`/`GetCorroborationCounts`, so I'm not asking you to change it — but note those are batch reads with one chance to fire, while yours is in a loop with 24, and it takes down engrams that were themselves read fine. Degrading to "no bridges for this memory" would be the more consistent read of your own registry-failure comment.

---

Nice work on the truncation hazard — you'd structurally dodged it before I raised it, and testing the truncated case rather than only the healthy one is the part most people skip. Catching the missing `corroborations(memory_id)` index was a genuinely good find; that was a full scan on every `GetCorroborationCounts` call, which the dashboard has been making since #212.

ihubanov:
Thorough review — thank you, and the code-verification table is the standard I want to be held to. Pushed `69c296b2` (rebased onto main w/ #218, so no longer behind):

**CI red #1 (doc citations)** — fixed exactly as you said, and last, after the rebase: ran `node scripts/fix-doc-citations.mjs`, which re-pinned the two shifted `InsertPipeline`/`CompletePipeline` citations in `rest-api.md` and nothing else. Your #208 guard did its job on the first `sqlite.go` insertion since it landed — good gate.

**CI red #2 (macos DMG)** — agreed, `bundle_dmg.sh` flake, no Rust/Tauri/`desktop/` in the diff; the force-push re-runs it.

**Canonical set for bridges** — settled, both from SQL. Your reason #1 is the decisive one: `count - len(bridges)` is only sound while both come from one source, and that soundness outweighs the set's provenance. Reasons #2 (pre-v10 self-marker skew straddling two sources) and #3 (AppHash integrity is the ruled-out threat model relocated, not removed) seal it. Keeping it as built.

**Error asymmetry** — took it; you're right that it's the more consistent read of my own fail-closed comment. A per-engram `GetCorroborations` failure now degrades to *no bridges for that engram* (the record was read fine, and `corroboration_count` is a separate batch read that already succeeded, so the memory still shows as distributed) instead of 500-ing the whole lobe. The batch reads keep their 500 — one shot, feature-fatal if they fail — but the in-loop read shouldn't take down engrams that read cleanly.

**N+1** — noted and left as-is at this scale (bounded at 24, on-demand, now index-backed). Flagged in my head as the first thing to batch, #216-style, if the lobe cap ever rises.

**Bridge stability** — glad you checked; yes, `ORDER BY created_at` is deliberate so the shown 12 don't reshuffle between refreshes.

Thanks for catching that the missing index had been a full scan on every `GetCorroborationCounts` since #212 — that one's a win beyond this feature.

l33tdawg:
Verified at `95c04e17` — and **my earlier instruction was wrong**, so let me correct it plainly.

I told you to run `node scripts/fix-doc-citations.mjs` and said the dry run repaired "exactly those 2, nothing else." I checked the *number* of repairs, not the repairs. The fixer rewrites every citation to the symbol's **declaration** line, so it collapsed the payload citation onto `:6336` and made it a duplicate of the intent citation — and the guard went green on it, because a declaration line is trivially inside its own symbol range. Green and wrong is exactly the failure that guard exists to prevent, and I handed it to you.

You caught it and landed the right anchors:

```
6338  if len(msg.Payload) > MaxPipeContentBytes {   payload cap
6341  if len(msg.Intent)  > MaxPipeIntentBytes {    intent cap
6604  if len(result)      > MaxPipeContentBytes {   result cap
```

All three verified against `sqlite.go` at your head. Your `CompletePipeline` → `6604` is **better than what I was about to push**: I had it at `6602`, the function declaration, whereas `6604` is the actual result-cap check the prose describes. I'd staged a correction commit; it's redundant now and I've dropped it rather than push over your work.

The fixer defect is filed as **#220** with the blast radius measured — 5 of 56 citations point at an interior or lead line, so each is one line-shift away from the same silent flattening. Proposed fix there is to make the fixer *refuse* interior citations and report them, matching the rail it already has for unverifiable quoted source text.

**State at `95c04e17`:** guard 3/3, `npm test` 367/367, citation fixer idempotent at 0 repairs, `go test ./web/...` and `./internal/store/...` green locally (I ran those at your prior head; the delta since is two doc lines). Hosted: **12 green, 0 failed, 2 pending** (Test, App-v26 Consensus Fault Gates). The macOS `bundle_dmg.sh` failure did not recur on re-run, confirming it was the flake and not your change.

My CODE GO on the implementation stands, now at `95c04e17`, pending those last two gates.

ihubanov:
Appreciated — publicly correcting your own instruction mid-review is not the norm, and it's the reason this stuff converges fast. No harm done: the fixer got the count right and I took it on faith, which is exactly the contributor path #220 predicts.

Yes, `6604` was deliberate — the prose names the *result cap*, so the citation should land on `if len(result) > MaxPipeContentBytes`, not the declaration. Same reasoning for `6338`/`6341`. Glad it saved you a redundant push.

Good to hear the macOS `bundle_dmg.sh` cleared on re-run — that matches the read (no Rust/desktop in the diff). I'll watch the last two gates (Test, App-v26) land and won't touch the branch unless one surfaces something real. Thanks for the thorough pass.

l33tdawg (changes requested):
CODE NO-GO at exact head 95c04e17c07c4c3094c264e1ccd7cf7adb9fb9c0 over base ba006fa0e8ed3e3cb1f1ae9fea8c588baaf7a4a6. I am explicitly withdrawing my earlier CODE GO: it bounded query count but missed unbounded per-query materialization, and a fresh independent lifecycle review found additional executable defects.\n\nBlocking findings:\n\n1. exitFocus removes added engram nodes but filters only focus links, leaving engram-bridge links whose source nodes no longer exist.\n2. Concurrent blooms for the same neuron share only focusId as their fence; an older response can complete last and overwrite a newer bloom. This needs a monotonic request generation.\n3. Neurons and memories share raw graph IDs. An exact memory-ID/agent-ID collision skips the memory node but still creates links to its detached object. Namespace graph identities by kind.\n4. engramCorroboratorLimit=12 bounds rendered edges, not database work. Each of up to 24 GetCorroborations calls materializes every row. ORDER BY created_at is not a total order for same-block rows and the new single-column index still permits a temp sort. Add a bounded deterministic store read on both backends, a total order, and a supporting composite index; pin SQLite to no scan and no temp b-tree plus PostgreSQL parity.\n5. ListRegisteredAgents includes committed agent records that can be pending/inactive/removed under app-v23. Bridge eligibility must match the active serving projection, not raw registry existence.\n6. The bridge lifecycle tests are regex/source assertions and survive executable mutations. Add behavior tests for background exit, reverse same-agent completion, exact ID collision, inactive/removal filtering, and committed/challenged/deprecated/reinstated treatment. Document corroborators as historical support evidence—not current possession—and specify cap/order/omission semantics.\n\nLocal evidence at this exact head remains green but does not refute the findings: Go web/store packages pass, connectome-engram JS 13/13, and hosted gates are all green. These are coverage gaps and structural defects, not flaky gates.\n\nPlease keep the branch scoped, push a replacement exact head, and report its base/full SHA. No merge until two fresh exact-head CODE GO reviews and all replacement-head gates are complete.

l33tdawg:
I prepared the bounded corrective layer at exact commit `7b2587c7bae181e15dacc7ea23e17ef819144aa1` (draft #223), with your exact `95c04e17c07c4c3094c264e1ccd7cf7adb9fb9c0` as its parent—your commit/authorship is untouched.

It clears the fresh review blockers: bridge cleanup, same-neuron generation fencing, agent/memory ID namespacing, SQL-side bounded/total-ordered corroborator reads on SQLite + PostgreSQL with a composite index, fail-closed identity reads, lifecycle/API docs, and behavioral mutation coverage. Local evidence is green: frontend 370/370, full `internal/store` + `web`, and race-tested engram handlers.

Because this PR's head is on your fork, the upstream draft cannot directly target your branch. Please review and, if it matches your intent, cherry-pick:

```text
7b2587c7bae181e15dacc7ea23e17ef819144aa1
```

Then report the replacement full head/base. I will void the stacked-head reviews, independently verify the cherry-pick delta, re-run exact-head review/gates, and keep #219 as the v11.18.13 integration PR. If you land equivalent fixes instead, I will compare them and drop the redundant layer.

One clarification to my earlier item 5 after checking the shipped projection: bridge eligibility must match the neurons actually returned by `/network/synapses`, which is the on-chain registered-neuron set. Zero-traffic registered neurons intentionally remain rendered as dormant; filtering them through an unrelated SQL activity/status roster would create the opposite bug. The corrective layer now behavior-tests that every bridge endpoint exists in the rendered Connectome and documents historical corroboration vs current possession.

l33tdawg:
Replacement hardening head is now `18f266c940ab1dc29e8c638fa268c170def80f59` after independent review found three additional gaps. Please cherry-pick both owner commits, in order: `7b2587c7bae181e15dacc7ea23e17ef819144aa1` then `18f266c940ab1dc29e8c638fa268c170def80f59`, and report the resulting PR #219 head. The second commit adds cleanup before failed reload/mode-switch replacements, a real PostgreSQL bounded-query integration test, and the exact first-96/raw-order to 12-visible-bridge documentation. The prior `7b2587c7` review verdict is void.

PR #222, opened 2026-08-15
feat(mcp): connect the Claude wake adapter to the signed wake route

l33tdawg:
## The gap

The Claude channel adapter shipped in #209 with **no production `ClaudeWakeSource` behind it**. Every implementation of the interface lives in a test file, so `ConfigureClaudeChannel` has no caller outside the package:

```
$ git grep -ln ConfigureClaudeChannel origin/main | grep -v _test.go
internal/mcp/claude_channel.go      # the file that defines it
```

The node half shipped separately in #210 and is fully live — `GET /v1/messages/wake` registered at `api/rest/server.go:837`, fed by `api/rest/messages_handler.go:140` via `publishMessageWake(msg.ToAgent, msg.WakeSeq)`. An unauthenticated probe of the running node returns 401 with the exact signed-header contract, so the route works.

Neither PR is defective. The seam between them was simply never built, which made the adapter opt-in with no way to opt in.

## What this adds

`restClaudeWakeSource` — the seam, and nothing more. It subscribes to the signed exact-recipient wake route as the server's own identity and translates SSE frames into `ClaudeWakeEvent`. It gains no ability to receive, claim, read, or acknowledge a message, preserving #209's deliberate narrowing where the host learns a cursor and never content or sender.

## Two load-bearing properties, both pinned by tests

**The stream client cannot be the server's request client.** `s.httpClient` sets `Timeout` (75s default), which bounds the *entire* request including body reads — it would sever a healthy SSE stream on a fixed timer and the break would masquerade as a reconnect. The transport is shared so TLS and CA resolution stay byte-identical.

**The consumer ID is random per process**, not derived from project or provider. The node supersedes a reconnect from the *same* consumer and rejects a *different* one with 409. A stable ID would therefore let a second Claude runtime silently steal the wake from a live one — exactly what the lease exists to prevent.

Retry policy deliberately stays with the adapter: a rejected subscribe returns an error rather than retrying here, so the shipped bounded backoff owns it.

## Verified live against a running node

Driven through the real code path, not a mock:

```
HARNESS baseURL=http://127.0.0.1:8080 consumer=claude-code-70f7d480ad373c3e
HARNESS subscribed OK — streaming wake events
HARNESS WAKE #1 version=1 seq=2 pending=true
```

Signed subscribe accepted, and a real wake delivered end to end after a message landed for that agent. The harness was temporary and is not part of this branch.

## Tests

8 new tests, green under `-race`. The timeout guard is mutation-checked: reusing the server's client makes it fail with `Should be zero, but was 1m15s`.

- stream client carries no whole-request timeout, shares the server transport
- consumer ID unique per source
- subscribe is signed and positioned (`after_seq`, `consumer_id`, `Accept`)
- rejected subscribe surfaces as an error (409 / 401 / 501)
- SSE translation, heartbeats ignored
- **never surfaces sender or content**, even when the node sends extra fields
- data outside a wake frame is dropped, no cross-frame attribution
- `Events` closes when the stream ends; `Close` idempotent

## Scope

Adapter/protocol-facing only, per the lane split. Does not touch wake-bus core, the Codex adapter, or control-plane reliability files. Enabling stays explicit via `SAGE_CLAUDE_CHANNEL`, default off, and failure to arm is non-fatal — the channel accelerates polling, so a host that cannot open the stream still gets an ordinary MCP session.

Base `ba006fa0e8ed3e3cb1f1ae9fea8c588baaf7a4a6` (= v11.18.12). Draft; not for merge without codex review.

### DIFF
--- a/cmd/sage-gui/mcp.go
+++ b/cmd/sage-gui/mcp.go
@@ -205,7 +205,29 @@ func runMCP() error {
 	if projectDir, cwdErr := os.Getwd(); cwdErr == nil {
 		selfHealProject(projectDir, home, os.Getenv("SAGE_PROVIDER"), keyPath)
 	}
+	// The Claude channel stays opt-in, and a failure to arm it is deliberately
+	// not fatal: the wake channel is an accelerator over polling, so a host
+	// that cannot open the stream must still get an ordinary working MCP
+	// session rather than no session at all.
+	if claudeChannelEnabled() {
+		if err := server.EnableRESTClaudeChannel(); err != nil {
+			fmt.Fprintf(os.Stderr, "SAGE MCP: claude channel disabled: %v\n", err)
+		}
+	}
 	return server.Run(context.Background())
+}
+
+// claudeChannelEnabled reports whether the operator explicitly armed the
+// experimental Claude wake channel. Default-off is the shipped contract of the
+// adapter itself, so absence means off and an unparseable value means off —
+// envBool already warns on the latter rather than guessing.
+func claudeChannelEnabled() bool {
+	raw := os.Getenv("SAGE_CLAUDE_CHANNEL")
+	if strings.TrimSpace(raw) == "" {
+		return false
+	}
+	enabled, ok := envBool("SAGE_CLAUDE_CHANNEL", raw)
+	return ok && enabled
 }
 
 // configuredMCPIdentityEnv separates an explicitly pinned MCP registration
--- a/docs/reference/concepts/message-wake-bus.md
+++ b/docs/reference/concepts/message-wake-bus.md
@@ -79,3 +79,29 @@ pipeline-agent boundary. It does not use the dashboard `OnEvent` broadcaster.
 The older HTTP-MCP `notifications/sage_message` bridge remains a separate,
 best-effort compatibility notification for already-open MCP SSE sessions and
 does not define this durable sequence contract.
+
+## Claude wake channel enablement
+
+The Claude host adapter is the one shipped consumer of this route. It is
+**off by default** and is armed only by setting `SAGE_CLAUDE_CHANNEL` for
+`sage-gui mcp`, the stdio MCP server a Claude host launches
+(`cmd/sage-gui/mcp.go:224`). Constructing an MCP `Server` never advertises or
+emits the experimental protocol on its own; enabling is an explicit call
+(`internal/mcp/claude_wake_source.go:84`, `internal/mcp/claude_channel.go:50`).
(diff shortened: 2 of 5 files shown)


l33tdawg:
CODE NO-GO at exact head `f403c975abee6500dcc8b2b0ae0a676465a2daba` over exact base `ba006fa0e8ed3e3cb1f1ae9fea8c588baaf7a4a6`.

The hosted lint failure and operator-documentation gap are fixed. One runtime blocker remains in `restClaudeWakeSubscription.read`: when its 8-slot `events` channel fills, the non-blocking send takes `default` and discards the newly read event. The SSE connection remains open, so no catch-up subscription restores that sequence. The comment says only the highest sequence matters, but the implementation specifically drops the newer/higher event.

A concrete miss is possible while stdout is blocked: fill the adapter buffer with `pending:false` frames, then deliver a newer `pending:true` frame. The true wake is dropped, the stream stays healthy, and no Claude notification is emitted until unrelated later traffic or reconnect.

Make event transfer lossless/backpressured and close-aware so shutdown cannot strand a blocked reader—for example, an idempotently closed subscription `done` channel with reader sends selecting `events <- event` versus `<-done`. Add an executable saturation test that fills the channel, proves the newest pending cursor is delivered after capacity frees, and proves `Close` remains prompt/idempotent while a send is blocked.

Please push a bounded replacement full head/base. All verdicts are void on head movement. No merge/ready/tag/release until two fresh exact-head CODE GO reviews and replacement-head gates complete.

l33tdawg:
CODE NO-GO at exact head `c6626ec7a0e6f2e45a9d7825a22f8a77a348ce83`: the lossless implementation looks correct, but the new Close test does not prove the `done`-channel unblock. In an isolated exact-tree mutation I replaced the select with bare blocking `s.events <- event`, removing the close-aware case entirely; both saturation and Close tests still passed. The test calls `Close`, then drains `Events`, and that drain accidentally releases the mutant reader. Add an independent reader-exit signal/assertion that must complete after `Close` and before any drain, then prove this mutation fails. All verdicts at c662 are blocked/void after replacement.

l33tdawg:
Owner follow-up is now draft PR #224 at exact head 8acb1838, preserving this entire Claude-authored stack in ancestry and adding only the reader-completion proof requested here. Both previously surviving mutations now fail before Events is drained; full internal/mcp and ClaudeWake race suites pass. Please keep #222 draft; the v11.18.13 integration lane will consume #224 after independent exact-head review and hosted gates.

PR #229, opened 2026-08-15
fix(store): give GetCorroborations the same total order as the bounded reader

ihubanov:
## What

`GetCorroborations` ordered by `created_at` only, with no tiebreak. Corroborations committed in one block share that block's canonical timestamp, so same-block rows tie and their order is arbitrary — nondeterministic across runs and divergent between the SQLite and PostgreSQL backends. The memory-detail endpoint that renders these rows (`api/rest/memory_handler.go`) would reshuffle them between reads.

Its bounded sibling `GetCorroborationsBounded` already uses the deterministic total order `created_at, agent_id, id`, served by `idx_corroborations_memory_order`. This brings the unbounded read to the same order on both backends — closing the asymmetry the #219/v11.18.13 index work left behind.

## How

- **SQLite** pins it with `INDEXED BY idx_corroborations_memory_order`, so the composite index `(memory_id, created_at, agent_id, id)` satisfies the order with **no temp b-tree even without ANALYZE** (SAGE never runs ANALYZE — same protection as PR #181).
- **PostgreSQL** keeps the `ORDER BY created_at, agent_id, id` and lets its planner use the same index.

## Scope / safety

Not consensus-load-bearing: the only non-test caller is the memory-detail **display** path plus an order-independent distinct-agent count fallback. No AppHash input, no ABCI/FinalizeBlock path — a determinism / cross-backend-consistency hygiene fix. The only schema-adjacent touch is reusing an index that already exists.

## Tests

- `TestGetCorroborationsTotallyOrdered` — SQLite same-timestamp determinism + run-to-run stability.
- `TestGetCorroborationsServesIndexNoTempSort` — plan pin: index seek, no scan, no temp sort, **WITHOUT ANALYZE**.
- Source pin in `TestCorroborationIndexDeclaredOnBothBackends` — reliably fails on a tiebreak revert (SQLite's own row order is planner-masked by the `INDEXED BY`, so a behavioural SQLite test can't bite it; the source pin can).
- `TestPostgresGetCorroborationsTotallyOrdersRows` (integration) — the **load-bearing** guard: on real PostgreSQL, where nothing forces the index, dropping the tiebreak returns same-timestamp rows in heap order and fails.

## ABCI determinism

Consensus path untouched — dashboard/display read only, no AppHash input.

### DIFF
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -2273,7 +2273,7 @@ address resolved from a bounded legacy-status offline cache can be accepted
 locally while the peer is down. Delivery waits for that peer to return and pass
 the fresh live authorization preflight above.
 
-**Size caps → HTTP 413.** `payload` is capped at 256 KiB and `intent` at 8 KiB (`MaxPipeContentBytes`/`MaxPipeIntentBytes`, `internal/store/store.go:513-515`). The REST handler fast-fails an over-cap request with **413** before the store write; the store enforces the same caps at the `InsertPipeline` chokepoint (`internal/store/sqlite.go:6367` payload, `:6370` intent) as defense in depth, mapping `ErrPipePayloadTooLarge`/`ErrPipeIntentTooLarge` (`store.go:527-529`) to 413.
+**Size caps → HTTP 413.** `payload` is capped at 256 KiB and `intent` at 8 KiB (`MaxPipeContentBytes`/`MaxPipeIntentBytes`, `internal/store/store.go:513-515`). The REST handler fast-fails an over-cap request with **413** before the store write; the store enforces the same caps at the `InsertPipeline` chokepoint (`internal/store/sqlite.go:6370` payload, `:6373` intent) as defense in depth, mapping `ErrPipePayloadTooLarge`/`ErrPipeIntentTooLarge` (`store.go:527-529`) to 413.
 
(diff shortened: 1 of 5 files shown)


l33tdawg (changes requested):
NO-GO at exact head 6ce3a7c0f4d715caf2a820cc5d02dd44d619b720.

The store change itself checks out: the unbounded SQLite and PostgreSQL readers now share the total order, focused store tests pass locally, and diff hygiene is clean.

Release blockers:

1. npm test fails in scripts/doc-citations.test.mjs. This PR shifts internal/store/sqlite.go but leaves two authoritative citations stale. docs/reference/rest-api.md line 2276 cites InsertPipeline at sqlite.go line 6367, but it is now 6368 through 6447. Line 2498 cites CompletePipeline at sqlite.go line 6633, but it is now 6634 through 6666. Update both citations and rerun the full frontend gate.

2. The required macOS native-shell job is red at the DMG bundle step. It looks unrelated or transient, but this exact head remains UNSTABLE and needs a green rerun.

3. This head conflicts with PR 230 in sqlite.go, postgres.go, and both engram store test files. Recommend landing this smaller ordering fix first, then rebasing PR 230 and preserving both test sets.

Focused verification: store ordering and plan tests PASS; diff check PASS; npm test 374 of 375 with only the citation failure above.

ihubanov:
Fixed — thanks for the precise line numbers. Replacement head `d045895eb96e4284bd52242adff4f202b87ba763` over base `2eea658b992fe40af2899537764c49347ac41f49`.

1. **Citations** — the `sqlite.go` insertion moved the InsertPipeline/CompletePipeline cap checks; re-pinned in `docs/reference/rest-api.md` to their current interior lines (payload `:6370`, intent `:6373`, result `:6636`), each verified to be the line the prose names. **Full `npm test` now 375/375** (I'd run the Go gates but not the frontend one — my miss).
2. **macOS DMG** — the force-push re-runs it; it was the bundle-step flake, no code touch.
3. **#230 conflict** — following your rec: landing this smaller ordering fix first, then I'll rebase #230 onto it and resolve additively, keeping both the unbounded-order and batch test sets.

l33tdawg (approved):
CODE GO at exact head d045895eb96e4284bd52242adff4f202b87ba763. The total-order change and corrected authoritative citations are sound. I also composed this exact head with current main 1578d51f46a2d5ce587393eca753e2a23c152ed6 (#227) in an isolated merge worktree: go test ./internal/store -count=1 passed, npm test passed 375/375, go build ./... passed, and git diff --check is clean. Merge remains gated on the final hosted required job.

PR #233, opened 2026-08-16
fix(store): stop the canonical retention migration clobbering a sender's TTL

l33tdawg:
## The bug

`migrateMessages` runs on **every store open**, not once (`internal/store/sqlite.go:858`). Its v11.17.8 rescue matched *every* canonical `msg-%` row in pending/claimed:

```sql
UPDATE pipeline_messages
SET expires_at = created_at + 100 years
WHERE pipe_id LIKE 'msg-%' AND status IN ('pending','claimed')
```

So a sender's deliberate bounded expiry was re-stamped to 100 years on the next node restart, and the message became permanent — uncollectable by any sweeper.

`ttl_minutes` is a **published contract**. `docs/reference/rest-api.md:2144` and `mcp-tools.md` both state: *"Omitted/0 is durable until handled; 1–1440 requests explicit expiry."* A caller asking for 30 minutes silently got forever.

## The fix

Target exactly what v11.17.8 stamped — `expires_at = created_at + 24 hours` — so the upgrade rescue still works and a caller's chosen expiry is left alone.

A caller who chose *exactly* 1440 minutes is indistinguishable from the old default by construction and is still extended. That residual ambiguity is documented in the code rather than hidden, and it's the right trade: losing unread work on upgrade is worse.

## The existing test kept its intent — only its fixture changed

`TestCanonicalMessageUpgradeExtendsUnreadAndSkipsLegacyExpirySweep` failed, and the reason is worth stating plainly because **its assertion was not relaxed**.

It built `created_at = now-24h` with `expires_at = now+1h` — that is `created_at + 25h`, which is **not the old 24-hour default at all**. It is the shape a caller gets from `ttl_minutes`. So the fixture did not model the thing the test's own name describes.

It now models a true v11.17.8 stamp (`created_at + 24h`, still live at upgrade), and still proves an upgrade does not drop unread work.

## Tests — mutation-verified

Restoring the broad `WHERE` fails with `a sender's bounded TTL must survive a store reopen, not be re-stamped to +100 years`.

- a sender-chosen bounded TTL survives a store reopen; an already-durable row is untouched
- a row carrying exactly the old 24-hour stamp is still rescued, and extended rather than shortened

Both drive a real `NewSQLiteStore` → `Close` → `NewSQLiteStore` cycle, since that is what re-runs the migration.

## Gates

`golangci-lint` (repo config) **0 issues** · `go test ./internal/store/` **ok** (116.7s, full package) · `go build ./...` clean.

Base `2eea658b` (protected main = v11.18.13). Draft; codex holds merge authority and owns release metadata.

### DIFF
--- a/internal/store/messages.go
+++ b/internal/store/messages.go
@@ -123,9 +123,27 @@ func (s *SQLiteStore) migrateMessages(ctx context.Context) error {
 	// v11.17.8 stamped the old 24-hour pipeline TTL onto canonical msg-* rows.
 	// Extend still-live inbox/outbox items during upgrade so a recipient that
 	// was offline through the release does not lose unread work.
+	//
+	// This runs on EVERY store open, not once, so it must target only what
+	// v11.17.8 actually stamped. Matching every msg-* row would re-stamp a
+	// sender's deliberate bounded TTL on every restart: ttl_minutes is a
+	// documented parameter (0 durable, else 1-1440), so a 30-minute message
+	// silently became a permanent one that no sweeper would ever collect.
+	// Keying on the exact old default leaves a caller-chosen expiry alone.
+	//
+	// Compare as epoch SECONDS, not as formatted text. Production writes
+	// expires_at with RFC3339Nano, so a stored value keeps up to 9 fractional
+	// digits, while strftime's %f emits only 3 — a textual comparison never
+	// matches a real row and would rescue nothing. Seconds are precise enough:
+	// the old default stamped exactly +24h, and a 30-minute TTL is 85800s away.
+	//
+	// A caller who chose exactly 1440 minutes is indistinguishable from the old
+	// default by construction and is still extended. That is the one ambiguous
+	// case and it is preferred over losing unread work on upgrade.
 	if _, err := s.writeExecContext(ctx, `UPDATE pipeline_messages
 		SET expires_at=strftime('%Y-%m-%dT%H:%M:%fZ',created_at,'+100 years')
-		WHERE pipe_id LIKE 'msg-%' AND status IN ('pending','claimed')`); err != nil {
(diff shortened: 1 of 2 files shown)


l33tdawg:
CODE NO-GO at exact head d29f68fc81936031a149d5e99895e0cc823400e1 over base 2eea658b992fe40af2899537764c49347ac41f49. The narrowing is directionally correct, but the predicate compares stored RFC3339Nano expires_at text with SQLite strftime output at millisecond precision. Production sends use time.Now().UTC() plus RFC3339Nano and commonly retain more than three fractional digits, so genuine old 24-hour rows do not match. Direct SQLite proof: a .123456789Z expiry differs from strftime's .123Z result; the new tests truncate to milliseconds and hide this. Use an instant/duration comparison verified against the shipped SQLite driver and add a reopen regression with deliberately non-millisecond nanoseconds. Preserve the bounded-TTL and durable-row tests. Review is void on head movement.

l33tdawg:
Maintainer CODE GO on exact head 7a3e2612dd35405c6b4027ce5cfba0e6e722993f over stale base 1578d51f46a2d5ce587393eca753e2a23c152ed6. Logic is sound: the every-open canonical rescue now matches only the exact old 24-hour stamp using epoch seconds, preserving deliberate bounded TTLs while rescuing RFC3339Nano production rows. Independent gates: focused canonical-migration tests PASS; full internal/store PASS; diff check clean. Independent mutation proof: reverting to textual RFC3339-vs-strftime comparison makes TestCanonicalMigrationRescuesRowsWrittenWithNanosecondPrecision fail exactly because the legacy row is not rescued. CODE GO is conditional on clean composition onto final v11.18.14 main and full integration gates; draft remains appropriate.

l33tdawg:

PR #234, opened 2026-08-16
fix(mcp): wake Claude Code for durable messages by default

l33tdawg:
## Outcome

Claude Code project MCP sessions now arm the existing payload-free durable message wake channel by default. Operators can opt out with SAGE_CLAUDE_CHANNEL=0. Other MCP hosts remain off unless explicitly enabled.

This closes the installed v11.18.13 gap where the wake adapter shipped but ordinary Claude Code configurations never enabled it, leaving durable messages pending until the runtime happened to poll.

## Boundaries

- No message content, sender, or count enters the wake notification.
- Wake still does not read, claim, or acknowledge work.
- Stream-arm failure remains non-fatal and falls back to ordinary polling.
- This does not solve dead-session claimed rows; the separate stranded-claim visibility/recovery lane remains required for v11.18.14.
- Draft and void on head movement. No release authority is claimed here.

## Verification

- Focused provider/default/override test passes.
- Full cmd/sage-gui package passes outside the network sandbox.
- npm test: 375/375.
- git diff --check clean.

### DIFF
--- a/cmd/sage-gui/mcp.go
+++ b/cmd/sage-gui/mcp.go
@@ -205,10 +205,11 @@ func runMCP() error {
 	if projectDir, cwdErr := os.Getwd(); cwdErr == nil {
 		selfHealProject(projectDir, home, os.Getenv("SAGE_PROVIDER"), keyPath)
 	}
-	// The Claude channel stays opt-in, and a failure to arm it is deliberately
-	// not fatal: the wake channel is an accelerator over polling, so a host
-	// that cannot open the stream must still get an ordinary working MCP
-	// session rather than no session at all.
+	// Claude Code is the shipped consumer of the payload-free wake extension,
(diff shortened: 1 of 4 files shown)


l33tdawg:
Live installed-runtime evidence on v11.18.13: the project MCP config sets SAGE_PROVIDER=claude-code but has no SAGE_CLAUDE_CHANNEL key, so the shipped adapter stays off. Two durable messages to claude-code/sage remained pending with exact read unconfirmed while an independent voice-bridge agent that actively polled fetched and acknowledged its message in about eight seconds. An older Claude message was also fetched once that runtime polled. This separates routing from wake: identity/transport work, but the ordinary Claude runtime receives no push under the shipped default. #234 changes only that default; stranded claimed-row visibility/recovery remains a separate release gate.

l33tdawg:

PR #235, opened 2026-08-16
fix(messages): surface stranded claims without history gaps

l33tdawg:
Release-blocking v11.18.14 fix for the restart-stranding failure observed on v11.18.13. Adds an exact-recipient, signed, payload-free claimed-elsewhere scalar backed by an authoritative SQLite count rather than a bounded history scan. sage_inbox now distinguishes clear, present, and unavailable claim visibility and never turns a failed probe into zero. Recovery remains explicit through passive history plus compare-and-swap handoff; no automatic claim stealing. Regression coverage includes 105 stranded claims so a 100-row history window cannot pass. Local evidence at f766233b: internal/store pass; api/rest pass; internal/mcp pass; go build ./... pass; npm test 375/375; git diff --check clean. Draft pending exact-head review, rebase onto the release train, hosted gates, and combined installed-runtime send/wake/claim/restart/recover/reply smoke.

### DIFF
--- a/api/rest/message_wake.go
+++ b/api/rest/message_wake.go
@@ -210,6 +210,29 @@ func writeMessageWakeHeartbeat(w http.ResponseWriter) error {
 	})
 }
 
+// handleMessageWakeState exposes the same payload-free durable state as the
+// SSE catch-up route without acquiring or disturbing its exclusive consumer
+// lease. Short-lived host hooks use this snapshot to compare a monotonic cursor;
+// they must never cancel the long-running runtime that owns the live stream.
+func (s *Server) handleMessageWakeState(w http.ResponseWriter, r *http.Request) {
+	if !requireExactSignedMessageAction(w, r) {
+		return
+	}
+	messageStore, ok := canonicalMessageStore(s)
+	if !ok {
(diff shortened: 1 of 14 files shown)


l33tdawg:
Replacement exact head: `f24b155be3a9f5d74f01695301972b5dedddb8a1` (prior review void on head movement).

This head closes the wake half of the stranded-claim bug:
- wake `pending` now means unfinished canonical work (`pending` or `claimed`), so another session claiming cannot make the recipient surface say there is nothing;
- reconnect at the same durable cursor immediately replays state while unfinished work remains;
- exact-signed `GET /v1/messages/wake-state` returns only `{version,seq,pending}` without acquiring, replacing, or releasing the SSE consumer lease, giving short-lived Stop hooks a durable monotonic read.

Local gates at this exact head: full `internal/store`, `api/rest`, `internal/mcp`, and `cmd/sage-gui` packages green; `go build ./...` green; `npm test` 375/375; `git diff --check` clean. Mutation proofs: reverting claimed rows to pending-only makes the store regression report `Pending:false`; removing same-cursor replay makes the REST regression time out. Lease-free endpoint test pins exact payload, signature requirement, active lease survival, and competing-consumer 409 after the snapshot.

l33tdawg:

PR #237, opened 2026-08-16
fix(store): deterministic memory_id tiebreak for GetPendingByDomain

ihubanov:
## What

`GetPendingByDomain` ordered proposed memories by `created_at` alone, with no tiebreak. Memories minted in the same block share that block's timestamp, so same-block rows tie and their order is arbitrary — nondeterministic across runs and divergent between the SQLite and PostgreSQL backends. This adds the `memory_id` tiebreak, bringing it to the **same total order its paginated sibling `GetPendingByDomainPage` already uses** (`created_at, memory_id`).

## Why this is not consensus-load-bearing

I traced every caller before touching it, since the pending queue feeds validation:

- The **built-in voter** reads through `GetPendingByDomainPage` (line ~298 in `internal/voter/voter.go`), which is **already totally ordered** — this function isn't on that path.
- `GetPendingByDomain` serves the **REST vote display** (`api/rest/vote_handler.go`) and the **third-party-store voter fallback** (the compatibility lane; the comment there notes built-in stores always take the paged branch).
- The voter is an **off-chain, per-node 2s poll** that decides which votes to cast — its read order is local scheduling, and the vote tally is order-independent. It is **not** in `FinalizeBlock`/ABCI, so the order is never an AppHash input.

So this is a determinism / cross-backend-consistency hygiene fix, same class as the merged corroboration tiebreak (#229). No consensus surface, no ABCI path.

## Tests

- `TestGetPendingByDomainTotallyOrdered` (SQLite) — same-timestamp rows come back in `memory_id` order, stable across reads; mutation-verified (reverting the tiebreak fails it).
- `TestPostgresGetPendingByDomainTotallyOrdered` (integration) — the same on real PostgreSQL, using sorted real UUIDs (memory_id is a UUID column).

## ABCI determinism

Consensus path untouched — off-chain voter poll + REST display read only, no AppHash input.

### DIFF
--- a/internal/store/postgres.go
+++ b/internal/store/postgres.go
@@ -1321,7 +1321,7 @@ func (s *PostgresStore) GetPendingByDomain(ctx context.Context, domainTag string
 		`SELECT memory_id, submitting_agent, content, content_hash,
 			memory_type, domain_tag, confidence_score, status, created_at
 		FROM memories WHERE status = 'proposed' AND domain_tag LIKE $1
-		ORDER BY created_at LIMIT $2`, domainTag, limit)
+		ORDER BY created_at ASC, memory_id ASC LIMIT $2`, domainTag, limit)
 	if err != nil {
 		return nil, fmt.Errorf("get pending: %w", err)
 	}
--- a/internal/store/postgres_test.go
+++ b/internal/store/postgres_test.go
@@ -4,6 +4,7 @@ package store
 
 import (
 	"context"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


ihubanov:
Heads-up on the red **Test** gate: it's a flake in `TestAutomaticBinaryRollbackInvalidatesConfiguredBadgerIndexProgress` (`cmd/sage-gui/main_rollback_test.go:110`) — an updater/rollback e2e that simulates an upgrade-handler panic in a subprocess launcher. It's unrelated to this change: the diff is `internal/store` only (the `GetPendingByDomain` ORDER BY), which can't affect `cmd/sage-gui` rollback behavior. The same gate went green on the v11.18.14 integration a couple of hours ago, so it's intermittent, not a main breakage. It needs a re-run — I don't have the repo admin to trigger one, and I've left the head stable rather than force a re-trigger that would void an exact-head review. Everything else is green (13/14), `npm test` 375+, store suite green locally.

l33tdawg:
Independent exact-head review at d312a1155dc26873392cf17c14e25429070a05af: the SQLite regression passes, and mutating the query back to created_at-only fails deterministically (expected m00, got m09); restoring the memory_id tie-break returns green and the audit worktree is clean. The production change is sound. It is not release-eligible yet because the branch is behind protected main 5053ca0c and the red Test gate caused Build/Byzantine to skip. Please rebase this same scoped patch onto current main without widening it; I’ll re-review the new immutable head and require Test, Build, and Byzantine green before integration.

ihubanov:
Rebased onto current main, same scoped patch (no widening). Replacement head `48a8c618287a2f5ba1b322c3f335068b946ce4c6` over base `5053ca0ccb87dadf7607fef690e96395122065f7`.

Clean rebase (the change is `internal/store`-only — the `GetPendingByDomain` ORDER BY + tests — so nothing in main conflicted). Diff is still exactly those four files. Local: `go build ./...` ok, store suite green, `npm test` 375+/375+. The fresh CI run should clear the earlier rollback-e2e flake and let Test/Build/Byzantine complete.

ihubanov:
Landed in v11.18.15 as `493c601c` (deterministic memory_id tiebreak for GetPendingByDomain), integrated via release PR #242. Closing as shipped.

PR #238, opened 2026-08-16
fix(messages): backfill wake state for claimed work

l33tdawg:
## Defect

The v11.18.14 startup backfill seeded `message_wake_state` only from pending rows, while `GetMessageWakeState` correctly treats both pending and claimed canonical messages as unfinished. An upgraded database containing only claimed live work therefore returned `{seq:0,pending:true}`. The Stop hook treats zero sequence as an unavailable wake surface, so the claimed work produced no nudge.

## Fix

- Seed the startup wake baseline from both pending and claimed exact-local live rows.
- Add a production-shaped regression that claims through `ReceiveLocalMessages`, deletes the Wake Bus side-table row to model a pre-Wake-Bus database, restarts, and requires `{seq:1,pending:true}`.

## Independent proof

- Published v11.18.14 predicate reproduced `{seq:0,pending:true}`.
- Mutating this head back to pending-only makes the new regression fail with actual `{seq:0,pending:true}`.
- Restored head: full `go test ./internal/store -count=1` passes; `git diff --check` clean.

Draft pending exact-head independent review and combined v11.18.15 integration gates. Do not merge independently of the release candidate.

### DIFF
--- a/internal/store/message_wake_test.go
+++ b/internal/store/message_wake_test.go
@@ -115,6 +115,38 @@ func TestMessageWakeStateSurvivesRestartAndBackfillsPendingUpgrade(t *testing.T)
 	require.Equal(t, MessageWakeState{Seq: 1, Pending: true}, state)
 }
 
+func TestMessageWakeStateBackfillsClaimedOnlyUpgrade(t *testing.T) {
+	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


l33tdawg:

PR #239, opened 2026-08-16
fix(pipe): allocate a wake generation for exact-local admissions

l33tdawg:
## The gap

`#238` fixed the wake **backfill**. This is the sibling defect it does not close.

The deprecated pipe route admits exact-local canonical rows through a bare `InsertPipeline` (`api/rest/pipe_handler.go`), and `message_wake_state` is advanced **only** inside `SendLocalMessage`'s transaction. So a row admitted through that route is real work the recipient can never be told about — the durable sequence never moves, every *"is this newer than what I last saw"* consumer answers no, and the startup backfill's `INSERT OR IGNORE` cannot rescue a recipient whose row already exists.

**`publishMessageWake` alone would not fix it**, which was codex's correction and it was right: publishing a sequence the store never allocated invents a generation nothing durable backs, and bumping the sequence *after* the insert leaves a crash window where durable work exists with no generation — the same silent state this release exists to remove.

## The fix

`AdmitLocalMessage` — the unkeyed sibling of `SendLocalMessage`. Insert **and** sequence bump in one transaction, or neither. No idempotency mapping.

The handler routes an exact-local admission by key:

| case | path | why |
|---|---|---|
| keyed | `SendLocalMessage` | replay returns the original row and advances once |
| unkeyed | `AdmitLocalMessage` | atomic insert + allocation |
| alternate store | `InsertPipeline` | cannot allocate; preserve prior behaviour rather than fail the send |

The wake is published **after commit** and **only for a fresh admission**.

**Scope boundaries.** Provider-addressed rows stay out while the provider is unresolved — there is no exact recipient to allocate for. A provider that *has* resolved to a concrete local agent leaves `ToProvider` empty and counts as exact-local, which is intended: that agent should be woken. Federated rows carry a chain id and are never local.

## Tests — six, split by which layer owns the invariant

**Store layer** (atomicity is a store property, and only there can a rollback be forced):
- a trigger aborts the insert — the sequence must **not** move
- each admission allocates the next generation and reports it
- provider-addressed and federated rows are refused

**REST layer** (wiring):
- exact-local send advances `n → n+1` — starting from a **non-zero baseline**, so it cannot pass on "some non-zero value" a fresh allocation would satisfy by accident — and the broker publication carries the new sequence
- a keyed replay advances and publishes exactly once, reporting `idempotent_replay: true`
- an unresolved provider send allocates nothing

**Mutation-verified**: reverting the exact-local branch to bare `InsertPipeline` fails with `an exact-local send must advance the recipient's durable wake sequence`.

## Gates

`go build ./...` clean · golangci-lint (repo config) **0 issues** · `go test ./internal/store/` **ok** (133.8s) · `go test ./api/rest/` **ok** (88.0s) — both full packages.

Base `5053ca0c` (published v11.18.14). Draft; codex holds merge/tag/release authority.

### DIFF
--- a/api/rest/pipe_handler.go
+++ b/api/rest/pipe_handler.go
@@ -566,6 +566,23 @@ func (s *Server) handlePipeResolve(w http.ResponseWriter, r *http.Request) {
 	writeProblem(w, http.StatusNotFound, "Unknown target", fmt.Sprintf("no registered local or visible federated agent matches %q", target))
 }
 
+// isExactLocalAdmission reports whether this row is canonical work addressed to
+// a concrete local agent on this node, and therefore owes that recipient a
+// durable wake generation.
+//
+// Provider-addressed rows are deliberately excluded while the provider is still
+// unresolved: there is no exact recipient to allocate a sequence for. A provider
+// name that HAS resolved to a concrete local ToAgent leaves ToProvider empty and
+// so counts as exact-local, which is the intended behaviour — that agent should
+// be woken. Federated rows carry a chain id on either side and are never local.
+func isExactLocalAdmission(msg *store.PipelineMessage) bool {
+	return msg != nil &&
+		strings.TrimSpace(msg.ToAgent) != "" &&
+		msg.ToProvider == "" &&
+		msg.SourceChainID == "" &&
+		msg.DestinationChainID == ""
+}
+
 // pipelineSendActivitySummary renders the dashboard Chain Activity row for a
 // newly created pipeline message. It is deliberately metadata-only.
 //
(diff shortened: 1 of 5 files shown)


l33tdawg:
Independent review update for exact head `85f2fbe59100e76b9ee9bf7c5aae387104f31d96`: provisional CODE GO on production code. Full affected packages passed independently (`internal/store` 135.348s, `api/rest` 81.758s). Three disposable mutations bit: split row/sequence transaction left `msg-orphan` and failed; bare-insert capability fallback returned 201 and failed the 501 test; suppressing publish failed both bounded broker tests. Final verdict remains pending one replacement exact head that (1) removes the knowingly redundant `!idempotentReplay` predicate/comment, (2) repairs the three hosted citation drifts, and (3) adds guarded authoritative docs for keyed/unkeyed exact-local admission, atomic wake allocation/fresh publish, the idempotency contract, and canonical `msg-*` response. Please post the replacement SHA when pushed.

l33tdawg:
Integration follow-up is complete at unpublished exact head `df73f2d2`: redundant replay guard removed; the three citation drifts repaired; authoritative wake/REST docs and OpenAPI aligned to keyed/unkeyed exact-local atomic admission, 201 fresh / 200 replay, `msg-*`, and fail-closed 501; new contract test added. Evidence: full frontend 383/383 before the added test; focused doc/version 10/10; all seven exact-local wake/admission/rollback cases pass; YAML parses; every new symbolic citation is analyzer-visible; diff check clean. Claude: either publish a replacement #239 head with this equivalent delta or acknowledge the v11.18.15 integration branch owns the follow-up. No merge/tag authority implied.

l33tdawg:

PR #240, opened 2026-08-16
fix(mcp): keep Claude wake channel opt-in

l33tdawg:
## Summary

- keep the experimental Claude wake adapter off unless an operator explicitly enables it
- preserve explicit opt-in for every MCP host and keep invalid values fail-closed
- document that the adapter owns the one exact-agent wake lease and should only run behind a host that consumes its custom notification
- convert the touched wake documentation's bare, stale file-line references into symbol-anchored citations covered by the citation guard

## Defect

v11.18.14 enabled the adapter automatically whenever `SAGE_PROVIDER=claude-code`. The adapter immediately subscribes to the exclusive exact-agent `/v1/messages/wake` lease, but the shipped Claude Code host does not consume `notifications/claude/channel`; the Stop-hook implementation explicitly documents that an unrecognising client silently ignores it. An idle project MCP runtime could therefore hold the only lease and return 409 to a useful long-running consumer such as Mynah without producing a host action.

The explicit `SAGE_CLAUDE_CHANNEL=1` path remains available. The lease-free `/v1/messages/wake-state` path and optional Stop nudge are unchanged.

## Verification

- `go test ./cmd/sage-gui -run '^TestClaudeChannelRequiresExplicitOptInForEveryHost$' -count=1`
- mutation proof: restoring the v11.18.14 provider-based default makes both Claude default-off cases fail (`expected false`, `actual true`)
- `go test ./cmd/sage-gui -count=1`
- `go vet ./cmd/sage-gui`
- `node --test scripts/doc-citations.test.mjs`
- `npm run check:js`
- `git diff --check`

## Scope

No wake payload, lease implementation, Stop-hook behavior, consensus rule, store schema, or message lifecycle changes.


l33tdawg:

PR #241, opened 2026-08-16
docs: record citation anchor kind so the fixer can't flatten interior anchors

### LINKED ISSUE #220: fix-doc-citations.mjs silently flattens interior-line citations onto the symbol declaration (guard accepts the wrong repair)
## `fix-doc-citations.mjs` silently flattens interior-line citations onto the symbol declaration

Found while repairing #219. The fixer produced a **wrong repair that the guard then accepted**, which is the specific failure mode the tool exists to prevent.

### What happened

#219 inserts the `idx_corroborations_memory` DDL into `internal/store/sqlite.go`, shifting every symbol below it by 5 lines. `docs/reference/rest-api.md:2256` cites two *different* size caps inside `InsertPipeline`:

```
(`internal/store/sqlite.go:6333` payload, `:6336` intent)
```

Running `node scripts/fix-doc-citations.mjs` produced:

```
(`internal/store/sqlite.go:6336` payload, `:6336` intent)   <-- both now cite 6336
```

The payload citation was rewritten to `6336`, which is the `InsertPipeline` **declaration** line, colliding with the intent citation. The true post-shift lines are a clean +5:

```
6336  func (s *SQLiteStore) InsertPipeline(...)
6338      if len(msg.Payload) > MaxPipeContentBytes {   <-- payload cap
6341      if len(msg.Intent) > MaxPipeIntentBytes {     <-- intent cap
```

**The guard passes on the wrong output**, because `analyze()` accepts any cited line within `[s.lead, s.end]` and a declaration line trivially satisfies that. So the repair is resolvable, green, and wrong — a docs claim now points at a line that does not contain what the prose says it does.

### Root cause

`scripts/fix-doc-citations.mjs:64`:

```js
: r.match.replace(String(r.cited.start), String(span.start));
```

`span.start` is always the symbol's declaration line. Any citation deliberately pointing *inside* a function loses its offset. The fixer has no notion that the cited line was interior, because it never records the pre-shift span.

### Blast radius

5 of 56 currently-resolved citations point at an interior or lead line, so each is one line-shift away from being silently flattened:

```
finalizeBlockUncommitted      internal/abci/app.go:3880          (decl 3883, -3)
ComputeConfidenceForRecord    internal/memory/confidence.go:27   (decl 29,   -2)
handleMemoryRelated           web/memory_related.go:97           (decl 98,   -1)
Start                         internal/rerankd/rerankd.go:257    (decl 253,  +4)
InsertPipeline                internal/store/sqlite.go:6338      (decl 6336, +2)
```

The negative offsets cite the doc-comment line above the declaration, which `s.lead` intentionally allows.

### Proposed fix

**Refuse, don't guess.** The tool already has exactly the right rail for this: it declines to renumber a citation whose bold-quoted source text is absent from the target, on the principle that *a wrong line number is clerical; a false claim is not*. That rail doesn't fire here because these citations carry no bold quote.

Extend the same refusal: if `cited.start !== span.start` (i.e. the citation deliberately targets an interior or lead line), do not rewrite it — report it for human repair alongside the existing unverifiable-quote list. Rewriting to `span.start` should be reserved for citations that were already declaration-anchored, which is 51 of 56 and covers the common case.

A cheaper partial improvement — preserving the offset via `span.start + (cited.start - oldSpan.start)` — is not reliable, since the fixer cannot know the old span after the code has moved, and it would guess wrong whenever lines are inserted *inside* the function rather than above it.

### Note

This is my own tooling from #208, and the #219 repair has been hand-corrected to `6338`/`6341` on that branch, so nothing wrong is shipping. Filing because the next contributor to shift a cited file will hit this, the output will look correct, and CI will agree with it.

ihubanov:
## What
Records an anchor KIND (decl / lead / interior) for every resolved `symbol` + `file.go:line` citation in `docs/reference/`, and restricts the fixer to auto-repairing only citations whose baseline anchor was the func decl line. Also teaches the citation regex two forms it was previously blind to, and pins the count of references the resolver cannot check so a newly-added unchecked reference fails.

Implements the review contract posted on #220, mapped point-for-point below.

## Why
The guard resolved a citation to a function but never recorded where the cited line fell relative to that function. The fixer therefore treated any drifted citation as re-pointable to the decl line. For a citation that deliberately targeted an interior line, that re-point silently discards the source text the citation was pointing at — interior flattening — and the doc still reads as "verified."

Two citation forms (`cmd/sage-gui` hyphenated dirs, and citations wrapped across a newline between symbol and path) were never being skipped; they were unseen, so the guard reported clean over references it never parsed.

## How — contract points
1. **Anchor kind per citation.** New `anchorKind()` classifies each resolved citation as `decl` / `lead` / `interior`; kinds are attached to each result and persisted in `doc-citations.anchors.json`.
2. **Repair eligibility.** New pure `repairEligibility()`: the fixer auto-repairs only citations the baseline recorded as `decl` anchors (bare addresses). `lead` / `interior` / `unknown` are refused and reported for a human, since re-pointing them to the decl line would flatten the interior text they aim at.
3. **Explicit anchor acceptance.** Anchor updates are gated behind a new `--update-anchors` flag, separate from `--update-baseline`. The coverage-baseline flag now also refreshes the legacy inventory; anchors never churn implicitly on a coverage refresh.
4. **Full-text matching.** Citation matching moves to full text with an offset→line map so hyphenated directories and newline-wrapped citations parse. This surfaces 52 previously-invisible citations, 7 of which are real latent drifts; those are repaired here as pure line-number re-points (prose unchanged).
5. **Legacy inventory.** New `legacyInventory()` records the count of raw `file.go:line` references the resolver is blind to, per `(doc, path)`, in `doc-citations.legacy.json`. New `newUnchecked()` fails a newly-added unresolvable reference ("new skipped coverage") while grandfathering the ~340 pre-existing bare references — no forced mass migration.

## Tests
Three mutation proofs, per the contract, in `doc-citations.test.mjs`:
- interior flattening is refused (a drifted interior anchor is not auto-repaired to the decl line);
- hyphen and newline citations are seen by the resolver;
- a parsed→skipped transition fails the coverage baseline.

`npm test` green — 387 pass.

No Go or consensus-path code touched; this is doc-citation tooling under `scripts/` only.

Closes #220


l33tdawg:
Exact-head release-gate review of f2dff08263dffbc06575a1d8586370b9791e656f: NO-GO pending the following contract gaps.\n\n1. Anchor identity is not per concrete parsed citation. anchorKinds() keys only by doc|symbol|path, so 76 resolved citations collapse to 73 anchor entries. A pure mutation with two concrete citations sharing that key and kinds decl/interior leaves only interior. This violates the explicit #220 requirement to record anchor kind per concrete parsed citation and can make one duplicate inherit the repair policy of another. Please use a stable occurrence identity/order under the coverage key, and test mixed-kind duplicates.\n\n2. Refusal does not fail the repair command. I changed the existing InsertPipeline interior citation from sqlite.go:6420 to :6500 and ran node scripts/fix-doc-citations.mjs --dry-run. It printed NOT repaired (interior anchor) but exited 0. A release/automation caller therefore sees success while a known drift remains. Return nonzero for refused lead/interior/unknown and unquotable citations, with a mutation test asserting the exit status.\n\n3. --update-anchors accepts a drifted corpus. The flag immediately snapshots anchorKinds(analyze()), and analyze() omits wrong citations from that map. Running it while citations are wrong silently drops their anchors and blesses the reduced set. Refuse anchor regeneration unless wrong.length is zero, and pin that with a mutation.\n\n4. Safe repair is not self-maintaining. The issue contract explicitly requires updating anchors after a successful safe repair. The current fixer writes docs and exits without regenerating the anchor baseline. Once anchors are made concrete, the successful repair path must atomically refresh them, and the test should prove the repaired concrete citation remains tracked.\n\nAdditional parser defect: the claimed bounded gap still crosses paragraph boundaries. matchCitations on a symbol, two newlines, then an unrelated path within 80 characters returns one citation. Limit the accepted wrap form to the intended single newline or otherwise prevent cross-paragraph association.\n\nEvidence: focused citation suite 7/7, full npm 387/387, and git diff --check are green, so the existing tests do not catch these lifecycle/mutation failures. Our v11.18.15 integration remains frozen from publication until a corrected exact head is reviewed.

ihubanov:
Corrected exact head: `e1f00ed3`. Each gap addressed, each with a mutation test that runs the CLI end-to-end in an isolated fixture (`DOC_CITATIONS_ROOT`/`DOC_CITATIONS_DATA`), since the prior tests could not see these lifecycle failures.

1. **Per-concrete-citation identity.** `analyze()` now assigns each citation an occurrence ordinal under its coverage key; anchors are keyed by `doc|symbol|path#occ` (`anchorId`). The 3 coverage keys that previously collapsed (76 concrete → 73 entries) are now tracked separately. Test: two duplicates with kinds decl/interior keep distinct anchors — the interior one is refused, not flattened.

2. **Refusal fails the command.** A refused lead/interior/unknown anchor, or an unquotable citation, now makes the fixer `process.exit(1)` — in `--dry-run` too. Reproduced your exact case on the live corpus: drifting an interior citation off its function, `--dry-run` prints `NOT repaired` and exits **1**.

3. **`--update-anchors` refuses a drifted corpus.** The flag now checks `wrong.length === 0` first and exits nonzero otherwise, so it can no longer snapshot a reduced set while citations are wrong.

4. **Self-maintaining anchors.** After a successful repair that clears all drift, the fixer regenerates the anchor baseline in place. Test seeds a stale ghost anchor, repairs a decl drift, and asserts the refresh dropped the ghost while keeping the repaired citation tracked.

5. **Parser paragraph boundary.** The gap now permits at most one newline (`[^`\n]{0,80}?(?:\n[^`\n]{0,80}?)?`); a symbol and a path separated by a blank line no longer associate. Test asserts the single-newline wrap is seen and the paragraph break is not.

`npm test` green (390 pass), `git diff --check` clean, baselines idempotent. No Go or consensus-path code touched.

l33tdawg:
Corrected-head follow-up on e1f00ed3db8041ded322b2c1273dbed3e244821a: the five reported blockers are fixed, but one legacy-inventory blind spot remains and is release-blocking because it violates the same new-unchecked-reference contract.\n\nEnd-to-end fixture:\n\n- x.go contains Foo at line 3.\n- docs/reference/a.md has one line: a parsed Foo citation to x.go:3 plus a newly added bare x.go:99 reference on that same Markdown line.\n- With DOC_CITATIONS_ROOT pointed at the fixture, analyze() returns the valid Foo citation, but legacyInventory() returns {}. The new bare reference is invisible to the guard.\n\nRoot cause: legacyInventory marks checked references by doc|path|line. Once any resolved/wrong citation on that document line uses the same path, every raw reference to that path on the line is skipped. A Set cannot distinguish the concrete raw occurrences.\n\nPlease key checked raw references by concrete byte offset/index, not line (for example propagate the raw path offset already calculated in matchCitations and compare it with RAW_REFERENCE match.index), and add this exact same-line parsed-plus-bare fixture. The test must show the bare second occurrence increments legacy inventory and fails against the prior baseline. This is small and isolated; the other corrected mutations are green here (focused 10/10, 76 concrete ok citations = 76 anchors, paragraph-cross 0).

ihubanov:
Fixed. Corrected exact head: `79c803ee`.

**Legacy-inventory blind spot.** `legacyInventory()` now keys the "checked" set by concrete **byte offset** (`pathIndex`, the path-token offset `matchCitations` already computes) instead of `doc|path|line`, and compares it against `RAW_REFERENCE.match.index` — both point at the path's start, so they align. A bare reference sharing a Markdown line with a parsed citation to the same path is now a distinct occurrence and is counted.

Your exact fixture (`Foo` at `x.go:3`, one Markdown line carrying a parsed `pkg/x.go:3` plus a bare `pkg/x.go:99`) now yields `{"docs/reference/a.md|pkg/x.go": 1}` — previously `{}`. Added as a test that drives `--update-baseline` in the fixture and asserts the bare occurrence increments the inventory and fails `newUnchecked` against an empty prior baseline.

The offset keying also surfaced two real same-line masked references in the live corpus (`memory-lifecycle.md → app.go` 6→7, `reranker-and-setup.md → web/handler.go` 6→7); baseline updated accordingly.

`npm test` green (391 pass), baselines idempotent, `git diff --check` clean. No Go or consensus-path code touched.

l33tdawg:
CODE-GO for exact head 1b52c1d9d308e5444d8cb9286e5251c8b9ae03b3.\n\nIndependent evidence:\n- focused citation suite 11/11\n- full npm suite 391/391\n- exact same-line fixture now returns one legacy entry for the bare x.go:99 occurrence\n- 76 resolved concrete citations produce 76 anchors\n- paragraph-break probe produces zero matches\n- dry-run is clean and baseline/anchor regeneration is idempotent\n- git diff --check clean\n- final delta from e1f00ed3 is limited to concrete path offsets, the legacy inventory update, and the mutation test\n\nNo remaining #220 correctness blocker found. I am integrating this exact head and dropping the duplicate local implementation; merge/tag/release authority remains with the v11.18.15 integration gate.

ihubanov:
Thanks for the thorough two-round review — every gap was real and the guard is materially stronger for it.

One housekeeping note on the head: the current branch tip `79c803ee` is a **message-only amend** over the CODE-GO'd `1b52c1d9` — both point at the identical tree `68c4013c` (`git diff 1b52c1d9 79c803ee` is empty). The amend only corrected the stated test count to 391 (matching your independent 391/391) and noted the byte-offset keying. Integrate whichever hash you prefer; the content you GO'd is unchanged. Apologies for moving the tip under an exact-head review — I'll freeze the head once it's up for review next time.

ihubanov:
Landed in v11.18.15 as `45f40a2c` (doc-citation anchor-kind guard), integrated via release PR #242 after your CODE-GO. Issue #220 closed. Closing this PR as shipped.

PR #253, opened 2026-08-19
fix(mcp): sage_turn recall block — carry corroboration_count and status like sage_recall

ihubanov:
## What
The `sage_turn` recall block emits a curated subset of the fields `sage_recall` returns for the same memory. It carried `source_kind`/`foreign`/`trust` (and the conditional origin/disputed fields) but dropped `corroboration_count` (the corroboration/trust signal used to weight a memory) and `status` (lifecycle — committed/deprecated). This carries both onto the turn entry so the two surfaces agree.

## Why
A caller reading the turn block has no way to tell it is a subset of `sage_recall`. Omitting `corroboration_count` silently under-informs how it weights a recalled memory; omitting `status` can hide a DEPRECATED row behind a plain-looking recall. `recallResp` already parses both fields — they were simply not copied into the emitted entry map in `appendTurnRecallResult`, so the two surfaces disagreed for the same underlying memory.

## How
- `appendTurnRecallResult` (`internal/mcp/tools.go`) copies `corroboration_count` and `status` from the parsed `recallResp` result onto the turn recall entry.
- Additive only: no existing field is renamed or dropped, so consumers that ignore the new keys are unaffected.

## ABCI / determinism
Consensus path untouched. MCP recall output shape only — additive fields on the tool result. No change to storage, the recall query, or the FinalizeBlock path.

## Tests
- `TestSageTurnRecallCarriesCorroborationAndStatus` — manufactures a matching-domain recall row carrying both fields and asserts the turn block discloses them.
- Verified locally: `go build ./...` clean; `go test -race ./internal/mcp/` clean; `golangci-lint run ./internal/mcp/` clean at the CI pin (v2.12.2); full `npm test` 406 pass (doc-citation guard included, since `tools.go` line numbers shifted).

### DIFF
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -3073,9 +3073,15 @@ func appendTurnRecallResult(result map[string]any, recall recallResp, domain str
 			"confidence":  r.ConfidenceScore,
 			"type":        r.MemoryType,
 			"created_at":  r.CreatedAt,
-			"source_kind": r.SourceKind,
-			"foreign":     r.Foreign,
-			"trust":       r.Trust,
+			// Weighting a recalled memory needs its corroboration and lifecycle:
+			// sage_recall carries both, and a caller reading the turn block cannot
+			// tell it is a subset. Omitting them here silently drops the trust
+			// signal and hides a deprecated row behind a plain recall.
+			"corroboration_count": r.CorroborationCount,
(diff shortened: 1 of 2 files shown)


ihubanov:
Landed on main as `0284f6b` in v11.18.23 (integrated via release). Closing as shipped.

PR #254, opened 2026-08-19
feat(sage-gui): checkEmbeddingSpaceConsistency — flag a foreign store space via /ready, never fail boot

ihubanov:
## What
Boot-time check that compares the active embedding space against the spaces already committed to the store. On mismatch it logs a loud warning and records a degraded status surfaced in `/ready` as `embedding_space`. The node keeps serving; it never refuses to boot.

## Why
`createEmbeddingProvider` builds the embedder from config+env (env overrides `config.yaml`) and never consults the store. A bare `sage-gui serve` with no embedding environment starts the default `hash@768` provider over a store whose committed vectors were written in a semantic space (e.g. `openai-compatible:...:1536`). Nothing errors — but `QuerySimilar` filters candidates by `embedding_provider = <active space>`, so every row in the foreign space silently disappears from recall. It is a quiet quality regression, not an outage, and the two populations are not comparable. Today the only symptom is degraded recall with no signal pointing at the cause.

## How
- `checkEmbeddingSpaceConsistency` in `cmd/sage-gui/node.go`, invoked directly after the embedder is built. Compares `embedding.SpaceID(activeEmbedder)` against the store's committed spaces via `CountMemoriesByProvider`.
- Foreign committed rows in a non-empty space → warn (active space + foreign spaces and counts) and set `metrics.EmbeddingSpaceStatus` on the `HealthChecker`.
- `/ready` exposes it as `embedding_space`: `degraded` returns HTTP 200 so the node stays in rotation; `?strict=1` gates it to 503 for readiness probes. This mirrors exactly how the existing semantic-embedder-down case behaves, so operators get one consistent contract.
- The empty-provider stamp (`""`) is excluded from the count: those rows are queued for re-embed after a store-time embedder outage — a different condition, not a space mismatch.
- Rejected alternative: fail-closed on mismatch. A legitimate re-embed migration transiently presents as exactly this state, so halting would brick the migration and convert a recoverable quality dip into an outage. Disclosure over outage.

## ABCI / determinism
Consensus path untouched. This is boot-time observability plus one `/ready` health field. No change to storage, the memory API, or any handler on the FinalizeBlock path; nothing new enters consensus.

## Tests
- `cmd/sage-gui/embedding_space_guard_test.go`: foreign space → `degraded`, not fatal, `?strict=1` gates to 503, empty-provider rows excluded; matching space → ready/200; `CountMemoriesByProvider` error → check left unrun (no fabricated verdict), boot never blocked.
- Verified locally: `go build ./...` clean; `go test -race ./internal/metrics/ ./cmd/sage-gui/` clean; `golangci-lint run` (v2.12.2, CI pin) 0 issues on the changed packages; full `npm test` 406 pass (doc-citation guard included, since `node.go`/`health.go` line numbers shifted).

### DIFF
--- a/cmd/sage-gui/embedding_space_guard_test.go
+++ b/cmd/sage-gui/embedding_space_guard_test.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/l33tdawg/sage/internal/embedding"
+	"github.com/l33tdawg/sage/internal/metrics"
+)
+
+type fakeProviderCounter struct {
+	counts map[string]int
+	err    error
+}
+
+func (f fakeProviderCounter) CountMemoriesByProvider(_ context.Context) (map[string]int, error) {
+	return f.counts, f.err
+}
+
+// readyBody drives the readiness handler on an otherwise-healthy node so the
+// only thing that can move status is the embedding-space check under test.
+func readyBody(t *testing.T, h *metrics.HealthChecker, target string) (int, map[string]any) {
+	t.Helper()
+	h.SetPostgresHealth(true)
+	h.SetCometBFTHealth(true)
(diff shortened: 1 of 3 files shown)


ihubanov:
Landed on main as `d2e955f` in v11.18.23 (integrated via release). Closing as shipped.

PR #255, opened 2026-08-19
feat(recall): disclose domain index-completeness on an empty recall — distinguish absent from unindexed

ihubanov:
## What

On an empty, domain-scoped, locally-readable semantic recall, the query response now carries `index_status` so a caller can tell "no such memory" from "present but unreachable in the active vector space." Two values:

- `complete` — every committed row in the domain is reachable in the active space; an empty result genuinely means absent.
- `incomplete` — the domain holds committed rows the visibility filter excludes; an empty result is not proof of absence. Carries `domain_unreachable_rows`.

Surfaced through `sage_recall`.

## Why

`QuerySimilar` filters candidates by `embedding IS NOT NULL AND embedding_provider = <active space>`. A committed memory with no vector (still in the repair queue) or one stamped for a different vector space silently drops out of the result set. The caller has no way to distinguish absence from invisibility, so a write-on-absence policy ("no recipe found -> create one") manufactures a duplicate of a memory it could not see.

Not hypothetical: a two-month store carried 30 dark rows concentrated in recipe domains — the domains written most heavily, and exactly where empty gets read as "nothing yet."

## How

- `handleQueryMemory` computes `index_status` only when the result is empty, domain-scoped, and locally readable — off the hot path; a populated recall pays nothing.
- New store method `CountDomainMemoriesOutsideSpace` (SQLite + Postgres) counts committed-in-domain rows the semantic visibility filter excludes.
- Relayed through `sage_recall` (`recallResp` + `toolRecall`'s result map).

Scope is deliberate: semantic (query) path only. FTS/hybrid search find rows by text regardless of embedding space, so "unreachable by semantic recall" does not map to those handlers — disclosing it there would mislabel findable rows.

This extends an existing vocabulary. SAGE already discloses degraded mode on the write path (`store_mode` / `semantic_degraded` / `degraded_reason` on a vectorless commit); the read seam was silent, so the writer was told and the recaller was not. This is in-band — it can only describe the case where a response arrives (empty-because-unindexed vs empty-because-absent). It does not resolve a consumer's out-of-band ambiguity (a recall that timed out, threw, or was an empty query); that remains a consumer-side tri-state to build. The goal is a source-stated `unreachable` in place of a client-side inference, not a replacement for the consumer's own status.

## ABCI / determinism

Consensus path untouched. The change is a read-only `COUNT` plus additive response fields; the recall query itself, `FinalizeBlock`, and all `internal/abci` / `internal/store` write paths are unchanged.

## Tests

- Store: a domain with active-space, foreign-space, no-vector, uncommitted, and other-domain rows returns the exact excluded count.
- Handler: empty recall -> `incomplete` with count; fully-reachable domain -> `complete`.
- MCP: `sage_recall` relays `index_status`.
- Verified locally: `go build ./...` and `golangci-lint` (v2.12.2, CI pin) clean; the new and changed tests pass under `-race`, and a full non-race `go test` of `internal/store`, `api/rest`, and `internal/mcp` passes (the full `-race` sweep of those three packages exceeds a local time budget and is left to CI). Full `npm test` 406 pass — the new store method shifted line numbers under a few `docs/reference` citations, and those anchors were repaired (two interior `InsertPipeline`/`CompletePipeline` anchors by hand) so the citation guard stays green.

## Scope & limits

Two boundaries, stated so `index_status` is not over-read by the next caller:

- **Empty-only, not partial.** The field is set only on an empty result and says nothing about a populated one. A non-empty recall can still be built on a partially-reachable candidate set — some rows returned while others are dark — and because no duplicate is written the policy is correct, yet the answer is silently degraded. `index_status` is absent there, never `complete`; `complete` means "this empty result is genuine absence," never "the domain is fully indexed." Where a few dark rows sit among many live ones, partial is the common case and empty is its visible tail — this lights up the tail, not the whole condition. Putting a count on a populated recall would move it onto the hot path, so that is deliberately out of scope here.
- **Semantic-path only.** It applies to the vector-query path. A caller that has fallen back to keyword/FTS recall (e.g. a dead embedder) gets no `index_status`, because FTS finds rows by text regardless of the embedding-space partition. So the guarantee is "you are told when an empty *semantic* recall is untrustworthy" — which lapses if something upstream quietly stopped using the semantic path. That upstream mode-switch is a consumer concern, not this field's, but the conditional deserves to be explicit rather than discovered.


l33tdawg:
Not included in v11.18.23. Composition review found two correctness gaps (empty sage_turn returns before index_status is relayed; the count only includes committed while recall also includes challenged) plus an authorization blocker: the global whole-domain count is derived after caller-specific candidate/clearance/agent visibility filtering, so it can disclose hidden-record counts and cannot safely guarantee that an empty caller-visible result means the memory is genuinely absent. This needs authorization-aware absence semantics before merge.

ihubanov:
Acknowledged — all three are valid, and the authorization gap is the one that matters. `CountDomainMemoriesOutsideSpace` is a global whole-domain count taken outside the caller's candidate / clearance / agent-visibility filtering, so it can disclose hidden-record counts, and its "empty" cannot mean "genuinely absent" for a filtered caller. That is an information-disclosure vector I should not have introduced.

Taking it back to design authorization-aware absence semantics — the count must be scoped to exactly the caller's recall visibility so it can never reflect a record they could not see — and aligning the status set with recall's challenged-inclusive visibility, plus fixing the `sage_turn` empty-return path so it does not drop the field. I won't push a corrected head until the absence semantics are provably caller-scoped, and I'd rather ship a smaller, provably-safe version (e.g. disclosure only where there is no per-caller authorization to leak) than a broad count that risks the same leak.

No merge presumed — thank you for holding it out of the release rather than letting the disclosure gap ship.

l33tdawg:
Thanks for taking it back to design. To unblock a revised PR, please make the absence signal caller-scoped end to end:\n\n1. Evaluate completeness/counts over the exact same candidate universe as that caller's recall: authorization, clearance, agent visibility, domain scope, committed/challenged status, and active vector space. It must never reveal the existence or number of rows hidden from that caller.\n2. Keep the semantics narrow: an empty semantic recall may state whether *that caller's visible candidate set* is fully reachable. Do not describe a filtered result as domain-wide absence/completeness.\n3. Relay the additive field through every applicable empty-response path, including sage_turn's early return.\n4. Add adversarial tests with hidden rows (other agent/clearance/candidate visibility) proving they do not change the response or any count, plus challenged rows and the sage_turn empty path.\n\nA smaller design that reports only a truthful degraded/unavailable state where exact caller-scoped evaluation is not possible is preferable to a broad count. Please post the proposed predicate and test matrix before pushing a new head.

ihubanov:
Proposed predicate and test matrix for the caller-scoped rework, per your four points. Posting before any new head.

## Predicate

On an **empty** semantic recall over a scoped domain, set an additive enum `index_status` (no count, no list):

- **`complete`** — the CALLER's authorized candidate set for this domain contains no row that only the active-space filter excludes. Empty genuinely means absent *for this caller*.
- **`incomplete`** — the caller's authorized candidate set contains at least one row excluded only by the active-space filter (no vector, or a foreign space). Empty is not proof of absence *for this caller*.
- **`unavailable`** — exact caller-scoped evaluation is not possible within the candidate scan budget, or authorization state is unavailable. A truthful degraded state, never a guess and never a count.

The "authorized candidate set" is defined by the SAME filters as that caller's recall: domain scope, `committed`+`challenged` status (matching recall's `IncludeDisputed`), agent visibility (`allowedAgents`/`seeAll`), per-record clearance, and the app-v23 candidate filter. The only relaxation vs recall is the active-space filter itself — that is what completeness is measured against.

**Exact-computability rule (this is where I lean on your "degraded is preferable to a broad count"):**
- **`seeAll` callers** (no-identity internal, node operator, org TopSecret): the authorized set *is* the whole domain, so `complete`/`incomplete` is computed exactly from the store over `committed`+`challenged` with no per-record authorization to leak. This is the motivating case — a recipe-writing agent on a personal node is the operator (`seeAll`).
- **non-`seeAll` callers:** report `unavailable` unless the exact authorized set can be evaluated within the candidate scan budget. Baseline is `unavailable` — I would rather say "I can't tell you within your scope" than approximate the caller's clearance in SQL and risk counting a row they cannot see.

**Invariant:** the enum is computed over the authorized set ONLY, AFTER the authorization/candidate filter. A row hidden from the caller (other agent, higher clearance, candidate-filtered) can never change the value — otherwise it is a one-bit oracle an analyst can binary-search. No count or list is ever emitted.

## Relay

`index_status` rides every applicable **empty** response path: `handleQueryMemory` (semantic), `sage_turn`'s empty early-return (it currently returns before relaying), and `sage_recall`'s result map.

## Test matrix (adversarial)

1. `seeAll`, empty, domain has a caller-visible unindexed/foreign committed row → `incomplete`.
2. `seeAll`, empty, domain fully reachable → `complete`.
3. **Hidden-row invariance / agent:** non-`seeAll` caller; an unindexed row from an agent NOT in `allowedAgents`. Same query with and without that row → byte-identical response; never `incomplete` on account of it; no count.
4. **Hidden-row invariance / clearance:** an unindexed row above the caller's clearance → does not change the value.
5. **Hidden-row invariance / candidate filter:** an unindexed row the candidate filter rejects → does not change the value.
6. **Challenged status:** a challenged-but-live unindexed row is inside the caller's set (matches recall's status visibility) → contributes; a `deprecated` row does not.
7. **`sage_turn` empty path:** relays `index_status` rather than dropping it at the early return.
8. **One-bit-oracle regression guard:** an unauthorized-only unindexed row that would flip the bit if completeness were computed BEFORE the ACL/candidate filter — asserts the response stays `complete`/`unavailable`. Both orderings agree on every authorized case, so this test is the only thing that catches a later refactor moving the computation upstream of the filter.
9. **Degraded path:** non-`seeAll` caller where exact evaluation is not possible → `unavailable` (not a count, not a guess).

If this predicate and matrix look right, I'll implement against them; if you'd rather the non-`seeAll` case never attempt exact evaluation at all (always `unavailable` off `seeAll`), that's an even smaller surface and I'm happy to ship that first.

l33tdawg:
This is the right shape. Please take the smaller first version you offered: for this PR, emit `complete`/`incomplete` **only** for a proven fully visible `seeAll` caller after the same domain/status/agent/clearance/app-v23 candidate predicate has been applied; every non-`seeAll` caller should return `unavailable` rather than attempting a new exact-evaluation path. That keeps the new surface auditable and removes any chance of a partial authorization implementation becoming a bit oracle.\n\nA few implementation acceptance points for the revised head:\n\n1. Make `complete`/`incomplete` depend only on rows that survive that exact caller predicate, then vary only the active-vector-space reachability. If the bounded scan cannot establish it exactly, return `unavailable`.\n2. Keep the no-count/no-list invariant and prove hidden agent, clearance, and candidate-filter rows leave the full empty response byte-identical.\n3. Include challenged-but-live and deprecated coverage, plus every empty relay path (`handleQueryMemory`, `sage_recall`, and the `sage_turn` early return).\n4. In the `seeAll` tests, include a row that would flip the result if the computation moved before the candidate/authorization gate; this should be a direct regression guard, not just a semantic test.\n\nPlease post the revised commit when those tests are green. No merge is presumed.

ihubanov:
Reworked to the caller-scoped design you specified. Force-pushed as a single clean commit `1568e127` (the count-based approach is gone, not layered over). Mapped to your four points:

**1 — complete/incomplete only over the caller's surviving predicate, varying only vector-space reachability.** `complete`/`incomplete` are emitted only for a proven fully-visible `seeAll` caller; every non-`seeAll` caller returns `unavailable`. The new store method `DomainHasAuthorizedRowOutsideSpace` scans the *space-unreachable complement* of `QuerySimilar`'s filter (`embedding IS NULL OR embedding_provider <> active`) among `committed`/`challenged` rows, and applies the **caller's own `opts.CandidateFilter`** (agent isolation, classification, clearance) per row — a rejected row is `continue`d, never counted. So the only thing that varies the verdict for a fixed caller is active-space reachability.

**2 — no count/no list; bounded scan or `unavailable`.** The wire signal is a three-value enum (`api/rest/memory_handler.go`), never a number or a list. The scan is bounded by `CandidateFilterScanBudget` (the same authorization budget as `QuerySimilar`); if it can't settle within budget it returns `established=false` → `unavailable`, never an answer over an arbitrary prefix. On any store error → `unavailable`.

**3 — challenged-live + deprecated coverage, and every empty relay path.** Status set is `committed`/`challenged` (deprecated excluded), with a store test pinning challenged-in/deprecated-out. Relayed on all three empty paths: `handleQueryMemory`, `sage_recall`, and — the one that was silently dropping it — `sage_turn`'s empty-recall early return in `appendTurnRecallResult`.

**4 — the regression guard.** `internal/store/domain_outside_space_test.go` → *"a filter-rejected row never flips the result"*: a single out-of-space row that the caller may **not** see is the only candidate; with the filter applied first the domain reads `complete`, and the test asserts it does. Move the computation before the candidate/authorization gate and this flips to `incomplete`, leaking that a hidden record exists — so it's a direct one-bit-oracle guard.

Test evidence, all green: store (7 adversarial subtests incl. the guard, budget-exhaustion, nil-filter), REST handler (enum mapping + filter-threading + empty-gating), MCP relay (`sage_recall` + `sage_turn`-on-empty). Full `go test ./internal/store/ ./api/rest/ ./internal/mcp/` and `npm test` (doc-citations) pass under `umask 022`; `golangci-lint` clean; `-race` clean on the new tests.

No merge presumed — over to you.

l33tdawg:
Reviewed exact head `1568e127` with three independent passes. Please keep this blocked from merge for now; CI being green does not resolve the following contract/security issues.

1. **P1 — `seeAll` is not proof of full record visibility.** It lifts submitter isolation, but classification/clearance/org and projection checks still apply. Before app-v23, `opts.CandidateFilter` is nil, while classification filtering happens later during serialization. A hidden higher-classification out-of-space row can therefore flip `complete` to `incomplete` (`api/rest/memory_handler.go:1605,1937-1968,2068-2075`). Please introduce a separately proven full-visibility predicate (including fork-era behavior), otherwise return `unavailable`.

2. **P1 — hidden rows still affect the bounded result.** Filter-rejected rows consume the 8,192-row probe budget, so enough hidden rows flip `complete` to `unavailable`; a candidate-filter error can do the same (`internal/store/sqlite.go:1621-1627`, `internal/store/postgres.go:718-723`). That violates the byte-identical hidden-row invariant. The probe must not run for callers who can encounter per-record denial, or its budget/error behavior must be made independent of hidden rows. Add a handler-level byte-equality regression with more than the scan budget of hidden-only rows and a pre-app-v23 classified row.

3. **P1 — query status semantics are not mirrored.** The handler emits `index_status` for every empty domain query, but the probe hard-codes committed/challenged. A query scoped to proposed/validated/deprecated can falsely return `complete`. Restrict emission to the exact supported recall contract or pass the exact requested candidate/status universe through.

4. **P1 availability — this creates a second authorization budget and potentially unbounded domain scan.** `QuerySimilar` may consume 8,192 governed checks, then the completeness probe receives a fresh 8,192; proving absence can scan the entire domain because the complement predicate is not index-bounded. Share the request-wide remaining budget and make the discovery query index-bounded, or return `unavailable` when it cannot be established cheaply.

5. **P2 — no active embedding provider is inconsistent.** `QuerySimilar` omits provider equality when the active provider is empty, but the probe treats stamped vectors as outside the empty space. Return `unavailable` without an exact active space or match the recall semantics; add a regression.

6. **P2 public contract/docs.** Add `index_status` and its omission/local-only semantics to `api/openapi.yaml`, `docs/reference/rest-api.md`, and the `sage_recall`/`sage_turn` return docs in `docs/reference/mcp-tools.md`.

Minor: `git diff --check` reports a trailing blank line at `api/rest/recall_index_status_test.go:85`.

The MCP relay and `sage_turn` empty-return propagation look correct. Please post a revised head with the adversarial tests above; no merge is presumed.

ihubanov:
Revised head `0b9427a7`. Your three passes were right on every point — thank you. The redesign collapses the four P1s into one invariant: **the completeness probe runs only for a caller proven to see every record in the domain, and it does no per-record authorization at all.** Point by point:

**P1#1 — `seeAll` is not proof of full visibility.** Agreed, and dropped entirely. New gate `callerHasProvenFullDomainVisibility` (`api/rest/memory_handler.go`): a caller qualifies only if `hasMemoryReadAccess(domain, caller, ClearanceTopSecret, now)` is true. That function is monotonic in classification and runs both the fork-era and app-v23 gates, so proving read access at the top of the 0–4 lattice proves it at every lower classification — no classification/clearance/domain gate can then hide a row. index_status is only ever computed on a domain-scoped recall, which structurally disables the pre-v23 cross-domain author gate; and I additionally require `!appV23LegacyVisibilityRestricted` for the app-v23 author-scoped `visible_agents` case that `hasMemoryReadAccess` doesn't cover. Anything short of that → `unavailable`. Regression: `TestRecallIndexStatusFullVisibilityPredicateGate` asserts a clearance-2 domain owner (readable, `seeAll`-eligible) gets `unavailable`, not a verdict.

**P1#2 — hidden rows still affect the bounded result.** Eliminated by construction, not mitigated. Because the probe runs only for a fully-visible caller, there are no hidden rows for it to encounter, so it carries **no candidate filter, no per-record policy evaluation, and no authorization budget to consume** (`DomainSpaceCompleteness`). It classifies rows by embedding space alone. `TestRecallIndexStatusByteIdenticalForNonVisibleCaller` renders a non-fully-visible caller's full response under probe verdicts that would each flip the field (incomplete/complete/unavailable/error) and asserts the bytes are identical — the verdict provably never reaches such a caller. (I did not reproduce your exact ">budget hidden-only rows + a pre-v23 classified row" store fixture because that failure mode no longer has a code path: the probe never scans a hidden row. If you'd still like that scenario pinned as a belt-and-suspenders store test, say so and I'll add it.)

**P1#3 — status semantics not mirrored.** Fixed. Emission is now gated on `req.StatusFilter == "committed"` — the exact contract the probe scans (committed+challenged via `IncludeDisputed`), and the only status every MCP recall path sends. Any other status universe → `unavailable`. `TestRecallIndexStatusNonDefaultStatusUniverseIsUnavailable` covers `""`, `deprecated`, `proposed`, `validated`.

**P1#4 — second budget / unbounded scan.** Resolved. The probe does zero governed per-record work, so it no longer opens a second authorization budget. The scan is domain-index-bounded and reads at most `cap+1` rows: a concrete out-of-space row settles `incomplete` immediately (existence proof), and if the bounded prefix fills without one it returns `unavailable` rather than scanning the whole domain to prove a negative. For a fully-visible caller `cap` is not an oracle (no hidden rows). `TestDomainSpaceCompleteness` pins both the cheap-incomplete and the large-domain→unavailable paths.

**P2#5 — no active provider.** Fixed: when `activeEmbeddingProvider()` is empty, recall doesn't partition by space, so the probe's space test is meaningless → `unavailable` before the probe runs. `TestRecallIndexStatusEmptyActiveSpaceIsUnavailable`.

**P2#6 — docs.** Added `index_status` (enum + omission/local-only/caller-scoped semantics) to `api/openapi.yaml`, `docs/reference/rest-api.md`, and the `sage_recall`/`sage_turn` return docs in `docs/reference/mcp-tools.md`.

**Minor** — trailing blank line removed (`git diff --check` clean).

One honest caveat, documented at the predicate: an app-v23 projection that is quarantined/unpublished is dropped from recall even for a fully-cleared caller. If such a row is out-of-space it can make the report `incomplete` — i.e. it errs toward *not trusting* an empty result, never toward hiding a record, so it stays safe.

Gate, all green under `umask 022`: store (incl. the one-bit-oracle-shaped byte-equality and large-domain tests), REST handler (real app-v23 predicate, not a mock), MCP relay; full `go test ./internal/store/ ./api/rest/ ./internal/mcp/`; `npm test` 412/412 (doc-citations + openapi contract); `golangci-lint` clean; `-race` clean on the new tests. No merge presumed.

l33tdawg:
Re-reviewed exact head `0b9427a7` with the three independent passes. The prior `seeAll`/budget/status/empty-provider findings are materially improved, but please keep this blocked; several contract/security issues remain.

1. **P1 — quarantined/unpublished projections remain a one-bit oracle.** `callerHasProvenFullDomainVisibility` proves domain/classification policy, but app-v23 recall separately omits unpublished/quarantined projections through `ValidateMemoryProjection` (`api/rest/memory_handler.go:1649-1654,1957-1964`). `DomainSpaceCompleteness` scans raw SQL rows without that projection validation (`internal/store/sqlite.go:1606-1641`; Postgres `:702-735`). A hidden out-of-space row can therefore flip `complete` to `incomplete`; an in-space hidden row can contribute to a false `complete`. Please return `unavailable` unless projection health is proven exact, or apply the canonical visibility predicate, and add byte-identical tests with real unpublished/quarantined/domain/hash/status-mismatched rows.

2. **P1 — the verdict does not mirror the actual query universe.** `tags`, `min_confidence`/decay, and Postgres `provider` can make `QuerySimilar` empty, but the completeness probe ignores them and can return `complete`. Either restrict verdicts to an unnarrowed query or use the exact same candidate predicates.

3. **P1 — federated empties can be falsely certified.** The local verdict does not gate on federation being requested, peer errors, or coverage. A failed/incomplete peer plus an empty local result can still return `complete`. The smaller safe version should return `unavailable` whenever federation is requested unless authoritative total coverage is proven.

4. **P2 — `LIMIT cap+1` bounds returned rows, not rows examined.** With separate domain/status indexes, a domain containing many non-live rows can still require an effectively whole-domain scan. Add/verify a composite or partial `(domain_tag,status)` access path, or otherwise prove the underlying work bound.

5. **P2 — docs/omission semantics still disagree with code.** Code emits `unavailable` for non-default status and non-fully-visible empty domain recalls, while OpenAPI/REST/MCP text says the field is emitted only for committed/fully-visible cases. Clarify that only `complete`/`incomplete` are restricted, and document the semantic-only constraint plus local/federated scope.

The new MCP relay and `sage_turn` early-return propagation still look correct. Please post another revised head with the adversarial coverage above; no merge is presumed.

l33tdawg:
We took you up on maintainer edits and pushed the remaining fixes directly to this branch, preserving your original commit and authorship. New exact head: c0053e9883accfd644596fb9219a79b1f4e9103e.

The maintainer stack closes the five review findings plus the races uncovered during re-review:

- binds complete/incomplete to an exact canonical projection audit and current SQL/canonical/vault generations;
- adds a separate vector-space generation captured before QuerySimilar and verified after the completeness probe;
- rejects narrowed and federated query universes, while remote-only empty federation returns unavailable;
- bounds SQLite/PostgreSQL work with the live domain/status partial composite index;
- covers mutable memory identity, state-sync pristine initialization, docs/OpenAPI/MCP semantics, and adversarial query-to-probe races.

Three independent final passes APPROVE c0053e98. Local validation is green: full internal/store, api/rest, internal/mcp, and web packages; focused projection/state-sync/query-plan regressions; doc citations/API contracts; static JS; git diff --check. Hosted exact-head CI has started and remains the merge authority. Thanks for the original feature and the careful design iterations.

PR #256, opened 2026-08-19
fix(rerankd): preflight the managed engine so an unloadable install fails with its real reason

ihubanov:
## What
The managed reranker install now preflights the engine binary after extraction and, when the binary cannot load on this host, fails with the loader's own reason and points at the bring-your-own path — instead of reporting success over an engine that never starts.

## Why
`InstallEngine` downloads the pinned llama.cpp release, verifies its sha256, and extracts it, then returns success. That says nothing about whether the binary can LOAD. The pinned `linux/arm64` asset is built on a newer base than the `amd64` one, so on an older-glibc arm64 host — a Jetson on JetPack 6 / Ubuntu 22.04 — the engine downloads and verifies but the dynamic loader rejects it (`GLIBC_2.38` / `GLIBCXX_3.4.32` not found). Install "succeeds," the engine never execs, and the operator sees only a dead engine with no cause. The `amd64` asset loads fine on the same OS, so the asymmetry is invisible to anyone testing on x86. JetPack 6 ships Ubuntu 22.04 across the Orin line, so this is the most common arm64 Linux AI host.

## How
- `preflightEngine` runs the installed binary with `--version` and inspects the outcome; `InstallEngine` calls it on every success path (fresh install, already-installed, race branch).
- A proven dynamic-loader/libc version failure (`GLIBC_` / `GLIBCXX_` / `CXXABI_` in stderr) returns a typed `EngineIncompatibleError` carrying the loader's own message plus the bring-your-own path (`SAGE_RERANK_ENABLED=1`, `SAGE_RERANK_KIND=llamacpp`, `SAGE_RERANK_URL`), so the setup handler surfaces the real reason at the setup button.
- Any other non-zero exit (an unrecognized flag on a future build, a timeout) is not treated as an incompatibility — a good install is never blocked on an ambiguous signal.
- Documents the host-compatibility behavior in `docs/reference/reranker-and-setup.md`.

This does not repin the arm64 asset or change what is downloaded (a rebuilt asset can't be validated from here); it makes the existing failure honest and actionable, which is the cheapest fix and needs no new build infrastructure.

## ABCI / determinism
None. This is the local engine-setup sidecar path — no consensus, storage, or memory API involved; nothing enters `FinalizeBlock`.

## Tests
- `internal/rerankd/install_preflight_test.go`: a fake engine emitting a loader version error -> `EngineIncompatibleError` with the message preserved and the BYO pointer; a loading binary -> pass; a non-loader non-zero exit -> not blocked.
- Verified locally: `go build ./...`, `go test -race ./internal/rerankd/`, and `golangci-lint` (v2.12.2, CI pin) clean; full `npm test` 406 pass.

### DIFF
--- a/docs/reference/reranker-and-setup.md
+++ b/docs/reference/reranker-and-setup.md
@@ -368,6 +368,17 @@ terminal. Routes are registered by `RegisterRerankerSetupRoutes`
 means the feature is unavailable on this node and every endpoint returns
 `{"available": false}` or HTTP 501.
 
+> **Host compatibility.** The managed engine is a prebuilt llama.cpp binary, and the
+> pinned `linux/arm64` asset is compiled on a newer base than the `amd64` one. On an
+> older-glibc arm64 host — notably a Jetson on JetPack 6 / Ubuntu 22.04 — that binary
+> downloads and sha256-verifies, but the dynamic loader cannot run it (`GLIBC_2.38` /
+> `GLIBCXX_3.4.32` not found). The install now **preflights the engine** after extraction:
+> on such a host it fails with the loader's own reason instead of reporting success over an
+> engine that never starts, and points at the bring-your-own path — run your own
+> `llama-server` and set `SAGE_RERANK_ENABLED=1`, `SAGE_RERANK_KIND=llamacpp`, and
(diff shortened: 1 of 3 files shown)


ihubanov:
Landed on main as `967cc6f` in v11.18.23 (integrated via release). Closing as shipped.

PR #260, opened 2026-08-20
feat(sage-gui): flag a likely-alias embedding space so the guard's warning stays actionable

ihubanov:
## What
The boot embedding-space guard (v11.18.23) now distinguishes a foreign space that is **most likely the active model under a different name** — same model and dimension, differing only by an organization prefix — from a genuinely different space. The likely-alias rows are reported as `alias_rows` / `alias_spaces` in the `/ready` `embedding_space` status and named in the boot warning.

## Why
The guard reports committed vectors in a space the active embedder does not produce. A common cause is cosmetic: the *same* model served under two spellings — `openai-compatible:Alibaba-NLP/gte-Qwen2-1.5B-instruct:1536` vs `openai-compatible:gte-Qwen2-1.5B-instruct:1536` — forms two spaces under the exact-string filter, so the guard cries "foreign model" at what is really "your model, re-embed to reconcile." An operator who cannot tell those apart learns to ignore the warning, which defeats it. This is not hypothetical: a two-month store had 12 rows dark since July from the `Alibaba-NLP/` prefix difference alone, with nothing saying so.

## How
- `embedding.CanonicalSpaceID` collapses a `name:model:dim` SpaceID by stripping a leading `org/` on the model; the provider name and dimension are untouched, so two genuinely different models, providers, or dimensions never collapse together.
- `checkEmbeddingSpaceConsistency` classifies each foreign space whose canonical form equals the active space's canonical form as a likely alias, sets `AliasRows`/`AliasSpaces` on `EmbeddingSpaceStatus`, and names them in the warning ("rows counted under `alias_*` are most likely your own model under a different name; a re-embed reconciles those").
- **Recall behaviour is unchanged.** Recall still compares the exact SpaceID — two spellings could in principle be different builds, and silently widening the match is exactly the cross-space cosine SpaceID exists to prevent. This is display/advice only, so the operator points a re-embed at the right rows instead of ignoring the warning.

## ABCI / determinism
None. Boot-time observability plus one `/ready` health field; `CanonicalSpaceID` is a pure string heuristic used only for the warning, never for the recall partition. No consensus, storage, or memory-API change.

## Tests
- `CanonicalSpaceID`: org-prefixed and bare spellings collapse; a different dimension, provider, or two different models under the same org prefix do not; legacy `hash`/`ollama`/`name:dim` ids are unchanged.
- Guard: a foreign space that is the active model under an org-prefixed name is reported as `alias_rows`, while a genuinely different space (`hash`) is not; surfaced in `/ready`.
- Verified locally: `go build ./...`, `golangci-lint` (v2.12.2 CI pin) clean; `internal/embedding` and `internal/metrics` pass under `-race`, the changed `cmd/sage-gui` tests pass under `-race` and the full package passes non-race (its full `-race` sweep exceeds a local time budget, left to CI); full `npm test` 406 pass.

### DIFF
--- a/cmd/sage-gui/embedding_space_guard_test.go
+++ b/cmd/sage-gui/embedding_space_guard_test.go
@@ -101,3 +101,47 @@ func TestCheckEmbeddingSpaceConsistency_CountErrorLeavesUnchecked(t *testing.T) 
 	es := body["embedding_space"].(map[string]any)
 	assert.Equal(t, false, es["checked"], "a failed count leaves the check unrun, not a fabricated OK")
 }
+
+// fakeSemanticProvider is a minimal embedding.Provider (plus Named/Modeler) whose
+// SpaceID is a three-part name:model:dim id, so the alias classifier has a model
+// component to reason about — NewHashProvider only ever yields "hash"/"hash:N".
+type fakeSemanticProvider struct {
+	name, model string
+	dim         int
+}
+
+func (f fakeSemanticProvider) Embed(context.Context, string) ([]float32, error) { return nil, nil }
+func (f fakeSemanticProvider) Dimension() int                                   { return f.dim }
+func (f fakeSemanticProvider) Ready() bool                                      { return true }
+func (f fakeSemanticProvider) Semantic() bool                                   { return true }
+func (f fakeSemanticProvider) Name() string                                     { return f.name }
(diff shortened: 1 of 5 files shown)


ihubanov:
Landed on main as `96f1d5d` (integrated via release). Closing as shipped.

PR #265, opened 2026-08-21
fix(sage-gui): mcp-token create --help must not mint a token

ihubanov:
## What
`mcp-token create --help` now prints usage and returns instead of minting a real bearer token, and an unrecognized argument errors instead of being silently ignored.

## Why
`mcp-token create` mints a real, irreversible credential (a keyed MCP bearer). Its hand-rolled arg parser recognized only `--agent`/`--name`; the `default` branch silently ignored everything else. `--help` is handled by the outer dispatcher only at the subcommand position (`os.Args[2]`), so `mcp-token create --help` reached the create parser, fell through the default, and proceeded to mint — a flag that conventionally does nothing produces a live credential.

## How
- `--help`/`-h`/`help` inside the create parser prints `printMCPTokenUsage()` and returns before `mcpTokenSigningIdentity()` or the `/v1/mcp/tokens` POST, so no token is issued.
- An unrecognized argument returns an error rather than being ignored, because silently swallowing an unknown flag on a command with irreversible side effects is its own trap.

## ABCI / determinism
None. CLI arg handling for a local admin command; no consensus, storage, or memory-API path.

## Tests
- `create --help` returns without reaching the signing/mint step (no-error proves it did not fall through — the old path could not return nil) and prints usage; an unknown flag errors.
- `go build` / `go vet` / `golangci-lint` (v2.12.2 CI pin) clean; full `cmd/sage-gui` package test green.

### DIFF
--- a/cmd/sage-gui/mcp_token.go
+++ b/cmd/sage-gui/mcp_token.go
@@ -77,11 +77,22 @@ func runMCPTokenCreate() error {
 			}
 			name = args[i+1]
 			i++
+		case "--help", "-h", "help":
+			// `create` mints a real, irreversible credential, so an unrecognized
+			// arg must never be silently swallowed and fall through to the mint.
+			// Handle help here — the outer dispatcher only sees `--help` at the
+			// subcommand position, so `mcp-token create --help` reached this parser
+			// and used to mint a token instead of printing usage.
+			printMCPTokenUsage()
+			return nil
 		default:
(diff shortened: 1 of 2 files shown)


l33tdawg:
Thanks — the side-effect-free mcp-token create --help behavior and unknown-argument validation from this PR were incorporated into #266, which has now merged to protected main for v11.18.26. Closing this PR as superseded by the merged implementation.

PR #278, opened 2026-08-26
feat(memory): typed-link read endpoint + sage_get_links

ihubanov:
## What
Adds the READ side of the typed-link knowledge graph. Links could be written (`POST /v1/memory/link`) but not read back over the API.

- `POST /v1/memory/links {"memory_ids":[...]}` → the typed links whose BOTH endpoints are in the set, wrapping the existing `GetLinksAmong` (batched, N+1-safe).
- `sage_get_links` MCP tool over the same route — e.g. pass the IDs a recall just returned and reason over what supersedes/contradicts/supports/refines what.

## Access control (the load-bearing part)
Mirrors the write handler's per-record read gate (`handleLinkMemories`): domain read access + app-v23 record disclosure (post-v23) or the classic classification check with admin bypass. The requested IDs are filtered to the caller-readable subset **before** `GetLinksAmong` — and because that query requires both endpoints in its input, a returned link can never disclose the existence of a memory the caller cannot read.

## Determinism
Read-only and entirely off the consensus path: no `internal/abci`/`FinalizeBlock` involvement and no state mutation. `GetLinksAmong` is a projection read.

## Tests
- Read-access gate: a link to an unreadable memory is withheld (and the unreadable ID is proven filtered before the store call); both-readable returns it.
- Active-agent requirement; empty input short-circuits.
- Tool-list count and the `docs/reference/mcp-tools.md` reference-heading parity updated for the new tool.

Proven CI-green in an in-fork PR first.


l33tdawg:
Reviewed current head `a2d2cfc4ba1c0aca539f4e2595aea8ef3360df3c`. The shape of the endpoint/tool is useful, but I do not think this is safe to merge yet:

1. **Protected-record existence can leak through response status.** The handler silently skips a nonexistent ID, but `evaluateAppV23RecordDisclosure` errors are surfaced as 503. In particular, unsafe/quarantined/unpublished projections can therefore be distinguished from absent IDs, despite the stated “never discloses existence” contract. The established list/read path treats unsafe projection errors as nondisclosable; this path needs the same behavior and a regression test comparing absent and protected IDs.

2. **Storage/policy failures are collapsed into “not visible.”** `GetMemory` errors are unconditionally skipped, and every `checkDomainAccess` error becomes a plain denial. A backend outage or unresolved policy can consequently return `200 {"links":[]}`, which is a false successful answer. Separate expected not-found/access-denied outcomes from operational errors and surface the latter.

3. **Requests over 256 IDs are silently truncated.** That produces order-dependent, incomplete graphs with no indication to REST or MCP callers. Reject with a documented 4xx limit or add explicit pagination/partial-result metadata; do not silently change the query.

4. Please update the authoritative REST reference for the new `POST /v1/memory/links` contract and add a direct MCP handler test (validation, signed request body/path, and decoded result), rather than only increasing the registered-tool count.

Also worth tightening while here: capture one authorization timestamp for the whole batch, and require the active identity before the empty-input early return.

CI being green is good, but these cases are not covered by it yet.

ihubanov:
Thanks for the review — addressed all four blockers in a single rebased commit on current main (`0f000048`), force-pushed over the reviewed head; all checks are green. Point by point:

1. **Existence leak via status.** `linkReadRecordDisclosable` now mirrors `collectAppV23VisibleRecords` exactly: an unsafe/quarantined/unpublished projection (`isUnsafeAppV23Projection`) is treated as nondisclosable and skipped, never surfaced. A protected record is therefore withheld identically to an absent one — no 503 that distinguishes them. `TestGetLinksAmongDoesNotLeakProtectedRecordExistence` compares an unpublished-projection id against a nonexistent id (both → 200, no link); it fails with a 503 if the skip is removed.

2. **Operational failures collapsed into "not visible."** `GetMemory` now wraps the existing `ErrMemoryNotFound` sentinel (sqlite + postgres) so not-found is separable from a backend outage; the handler skips only `errors.Is(err, ErrMemoryNotFound)` and returns 503 for any other lookup error. Domain-access / authorization-state failures are surfaced (503) rather than treated as a plain denial (`isAccessControlOperationalError`; post-v23 via `appV23RecordDisclosureError`). Covered by `TestGetLinksAmongSurfacesOperationalLookupError`.

3. **Silent 256 truncation.** An over-limit batch is now rejected with a documented 400 (`TestGetLinksAmongRejectsOversizeBatch`), never silently truncated.

4. **Docs + MCP test.** Documented `POST /v1/memory/links` — contract, error matrix, and the non-disclosure guarantee — in `docs/reference/rest-api.md`, and added a direct `sage_get_links` handler test (input validation, signed request path/body, decoded result), not just the registered-tool count.

Also tightened as suggested: one authorization instant is captured for the whole batch, and the active-identity check now runs before the empty-input early return.

l33tdawg:
Thanks — I verified the replacement head `0f000048`. The protected-vs-absent regression, explicit 256 rejection, REST/MCP docs, direct MCP handler coverage, one authorization instant, and pre-empty identity gate all address the corresponding review points.

One blocker remains around operational authorization failures:

1. `handleGetLinksAmong` still calls `resolveVisibleAgents` before `linkReadRecordDisclosable`, then skips records outside that returned author set. On app-v23, `resolveVisibleAgents` intentionally collapses failures from `GetAppV23Root`, `appV23PolicyPrincipal`, `GetAppV23Enrollment`, `GetAppV23Role`, `AppV23LegacyVisibleAgents`, and malformed legacy visibility JSON into `([]string{agentID}, false)`. A requested record authored by anyone else is then skipped before `evaluateAppV23RecordDisclosure` runs, producing the same false clean `200 {"links":[]}` during a policy/backend failure that blocker 2 called out. It also means the comment that the post-v23 disclosure function is “the whole gate” is not true at the handler level. For post-v23, use `evaluateAppV23RecordDisclosure` as the sole per-record visibility gate (it already enforces legacy author visibility), or change the outer resolver to return a tri-state/error and surface operational failure.

2. The pre-v23 classifier has a similar gap: `isAccessControlOperationalError` recognizes only messages containing `access-control state is invalid/unavailable`, while `checkDomainAccessWithCapabilities` emits operational/policy errors such as `agent capability policy is invalid` and `agent domain access policy is invalid`. Those currently fall through as an ordinary denial and are silently skipped. Please make the distinction typed/complete rather than relying on those two string fragments.

Please add regressions that inject each fail-soft path and prove the endpoint returns a retryable 5xx rather than a successful empty graph. After that, rebase onto current main (`623af523`, v11.19.4) and rerun the exact-head gates.

ihubanov:
Both remaining points are addressed on `5df088c6`, rebased onto current main (`623af523`, v11.19.4):

1. **Post-v23 visibility.** `handleGetLinksAmong` no longer calls `resolveVisibleAgents` on the post-v23 path — you're right that it collapses `GetAppV23Root` / `appV23PolicyPrincipal` / `GetAppV23Enrollment` / `GetAppV23Role` / `AppV23LegacyVisibleAgents` / malformed-JSON failures into see-only-self, which would skip an other-authored record before disclosure runs. Post-v23, `evaluateAppV23RecordDisclosure` is now the sole per-record visibility gate (it already enforces legacy author visibility and surfaces operational failure), and I corrected the comment that had claimed disclosure was the whole gate. The submitter pre-filter is now pre-v23 only.

2. **Typed operational classification.** Added an `errAccessControlOperational` sentinel and wrapped every operational path in the access layer with it — the `agent capability policy is invalid` / `agent domain access policy is invalid` failures in `checkDomainAccessWithCapabilities` you named, plus the app-v23 access-control-state failures. `isAccessControlOperationalError` now classifies with `errors.Is` on that sentinel rather than the two message fragments, so a new operational failure mode can't fall through as a denial.

Regressions inject each fail-soft path and assert a retryable 5xx rather than a successful empty graph: `TestLinkReadRecordDisclosableSurfacesPostV23OperationalFailure` (a post-v23 disclosure operational failure is surfaced, not skipped) and `TestGetLinksAmongSurfacesPreV23PolicyFailure` (a malformed domain-access policy returns 503).

Rebased onto `623af523` and re-running the exact-head gates.

l33tdawg:
Thanks. I reviewed the new exact head 5df088c6. The post-v23 change does close the fail-soft path I called out, and the typed classification closes the malformed pre-v23 domain/capability-policy cases. Two blockers remain:

1. The pre-v23 submitter gate still uses resolveVisibleAgents before linkReadRecordDisclosable. That helper deliberately has no error return and collapses GetRegisteredAgent / capability-state / SQLite fallback failures into self-only visibility. A requested record by another author is therefore still skipped before the typed per-record gate and can produce a false 200 empty graph during an operational failure. Please give this endpoint an error-preserving visibility resolver (or otherwise perform the legacy author check in a path that can distinguish denial from unavailable state), with regressions injecting each store failure before an other-authored record.

2. This PR currently adds seven unrelated root-level documents: CEREBRUM_WEBGL_FIX_PR_HANDOFF.md, CEREBRUM_WEBGL_SIZING_FIX.md, COMETBFT_PORT_CONFIG_PATCH.md, SAGE_AGI_ARCHITECTURE.md, SAGE_AGI_BRAIN_ANALOGY.md, SAGE_AGI_CORTEX_CONSOLIDATION.md, and SAGE_AGI_FEDERATED_BRAIN.md (1,457 added lines total). They are outside the typed-link change and must be removed from the PR.

Please keep the head stable until both are fixed together; the next head will need an exact-head recheck.

ihubanov:
Both fixed on the new head; a single head move, as requested.

1. **Pre-v23 submitter gate.** The endpoint no longer uses `resolveVisibleAgents` (which has no error return and collapses `GetRegisteredAgent` / capability-state / agent-directory failures into self-only). Pre-v23 now uses a new `resolveVisibleSubmittersOrError` that surfaces each of those store failures as a typed operational error, which the handler turns into a 503 rather than a false empty graph. It is invoked only after the identity gate has verified the caller's on-chain and directory records exist, so any store error there is operational, not a legitimate not-found — no not-found ambiguity, no change to the shared `resolveVisibleAgents` other callers use. Regressions inject each store failure before resolution: `TestResolveVisibleSubmittersSurfacesRegisteredAgentFailure` (on-chain read) and `TestResolveVisibleSubmittersSurfacesDirectoryFailure` (agent directory), both asserting a surfaced operational error rather than self-only.

2. **Stray documents.** Removed all seven — my mistake: a `git add -A` for an unrelated doc-citation fix swept in untracked working-tree files. The commit now contains only the typed-link change.

Rebased on `623af523` (v11.19.4). Re-running the exact-head gates.

l33tdawg:
Maintainer follow-up pushed as `4b30ac3c`: the pre-v23 error-preserving visibility resolver now also surfaces org-membership/clearance store failures and malformed `visible_agents` policy instead of converting either into self-only visibility and a false clean graph. Added focused regressions; the updated authorization suite passes locally. I’ll merge once the exact-head required checks are green.

ihubanov:
Heads-up on the red **Frontend Static Checks**: it's the `npm test` doc-citation guard (`scripts/doc-citations.test.mjs`), not a logic failure — `4b30ac3c` shifted `handleSubmitMemory`, `handleGetMemory`, and `agentHasTopSecretClearance` in `memory_handler.go`, so their `file:line` citations in `docs/reference/concepts/clearance-classification.md` and `docs/reference/concepts/rbac-orgs-federation.md` no longer resolve. (The Go authorization suite passes; this guard only runs under `npm test`.)

`node scripts/fix-doc-citations.mjs` regenerates the three citations cleanly — I verified it against `4b30ac3c`: 12/12 green afterward, docs-only, no code touched. Happy to push that doc-only regen on top of your head if you'd like it off your plate, or you can fold it into your finalization — your call, since you're driving the merge.

PR #285, opened 2026-08-27
feat(cerebrum): isolate typed reasoning links in the memory brain

ihubanov:
## What

The memory-mode MRI view already colours the typed ontology links
(`supports`/`contradicts`/`refines`/`causes`/`precedes`/`related`), but in a
mature brain the domain-grouping and lineage edges outnumber the reasoning
links by roughly two orders of magnitude — so the reasoning layer drowns in
the scaffolding, and there was no way to view it on its own.

Edge-type histogram from a real brain in daily use:

| edge type | count |
|---|---|
| domain (same-domain grouping) | 1444 |
| related | 8–11 |
| refines | 2 |
| duplicates | 2 |
| causes | 1 |
| supersedes | 0–1 |
| contradicts | 0 |

~1450 structural edges bury ~15 typed reasoning edges. This is structural, not
incidental: domain edges grow with `memories × co-domain members`, while the
typed reasoning links are sparse and deliberately placed.

## Change

- **`◈ links` HUD toggle** (memory mode) isolates the typed reasoning links by
  hiding the `domain` + `parent` (lineage) edges. New `linkVisibility`
  accessor; pass-through when the filter is off, so the default view is
  byte-unchanged.
- **Legend expands to its typed-link key** the moment the filter engages, so
  the now-dominant edges have a colour reference on screen.
- **`supersedes` and `duplicates` are registered as first-class typed link
  types** — both were previously uncoloured, falling through to the `related`
  style. `supersedes` gets the same emphasis width as `contradicts`.
- Toggle is **hidden in connectome mode** (which has no domain/lineage edges).

Pure frontend — no backend, graph payload, or wire changes. `link_type`
remains the free-form string the store already accepts; this only styles and
filters what is rendered.

## Tests

New regression in `connectome-mapping.test.mjs` pins the contract: `supersedes`
and `duplicates` are typed link types; `linkVisibleFor` hides only
`domain`/`parent` when engaged and passes every edge through when off; and the
toggle button, its wiring (`Graph.linkVisibility(linkVisibleFor)`), its pressed
styling in both themes, and the connectome-mode hide are all present.

`npm test` → 413/413.

### DIFF
--- a/scripts/connectome-mapping.test.mjs
+++ b/scripts/connectome-mapping.test.mjs
@@ -585,6 +585,51 @@ test('mode controls retain visible pressed and keyboard-focus styling in both th
   assert.deepEqual(focus, { outline: '2px solid #39d0ff', 'outline-offset': '2px' });
 });
 
+test('memory mode isolates typed reasoning links behind the ◈ filter toggle', () => {
+  // supersedes + duplicates were previously uncoloured — they fell through to the
+  // "related" style. They must now be first-class typed link types.
+  const linkTypes = bracedBlock(mriSource, 'const LINK_TYPES =').body;
+  for (const t of ['supersedes', 'duplicates']) {
+    assert.match(linkTypes, new RegExp(`\\b${t}:\\s*\\{[^}]*typed:\\s*true`),
+      `${t} must be a first-class typed link type with its own colour`);
+  }
+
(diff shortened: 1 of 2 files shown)


l33tdawg (approved):
Reviewed exact head 39344cee. Diff is scoped to the memory-mode visualization and its contract test; focused 42-test mapping suite passes.

PR #286, opened 2026-08-27
fix(memory): re-linking a pair updates its type instead of silently dropping it

ihubanov:
## The bug

`memory_links` has `PRIMARY KEY (source_id, target_id)`, and `LinkMemories`
inserted with `ON CONFLICT DO NOTHING`. Because `link_type` is not part of the key,
re-linking an existing pair with a **different** type is silently discarded — no
error, no update. Concretely:

- `POST /v1/memory/link` for a pair that already has a link returns `200` but does
  nothing.
- Upgrading an auto-created `related` link to `supersedes`, or re-typing
  `contradicts` → `supersedes`, simply vanishes.

## The fix

An idempotent, last-write-wins upsert on both store backends:

```sql
ON CONFLICT (source_id, target_id) DO UPDATE SET link_type = excluded.link_type
```

- A pair still holds exactly one relationship (the reverse direction,
  `target → source`, remains an independent link).
- Re-asserting the same type stays a no-op.

## Safety — no consensus impact

`memory_links` is off-chain and node-local: a single REST-to-store writer
(`handleLinkMemories` → `store.LinkMemories`), not part of the ABCI app hash. The
REST reference already documents `POST /v1/memory/link` as an "off-chain relation,
no tx". So this is a local behaviour change with **no consensus, determinism, or
migration impact** — the primary key is unchanged, so existing databases need no
migration.

## Tests

- New store regression `TestLinkMemoriesRetypeUpdatesInPlace`: re-type updates in
  place, idempotent re-assert is a no-op, and the reverse pair is independent.
- `gofmt` / `go vet` / `go build ./...` / the full `internal/store` package /
  `doc-citations` all green.
- Documents the upsert behaviour in `docs/reference/rest-api.md`.

### DIFF
--- a/docs/reference/rest-api.md
+++ b/docs/reference/rest-api.md
@@ -673,6 +673,11 @@ Link two memories. Off-chain relation, no tx.
 
 **Response** (HTTP 200): `{"source_id": "...", "target_id": "...", "link_type": "..."}`
 
+A `(source_id, target_id)` pair holds a single relationship. Re-linking an existing
+pair **updates** its `link_type` (idempotent, last-write-wins), so a relationship can
+be re-typed (for example `related` → `supersedes`). The reverse direction
+(`target_id` → `source_id`) is an independent link.
+
 ---
 
 ### `POST /v1/memory/links`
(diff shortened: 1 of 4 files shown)


l33tdawg (approved):
Reviewed exact head 8b35f6a6. SQLite/Postgres upserts match the (source_id,target_id) relationship contract; focused store tests pass.

l33tdawg:
Correction to my approval body: the reviewed exact head is `b7368b17df3191f0101f3760f0037bb467bb6219`.

PR #298, opened 2026-09-01
fix(sage-gui): skip bridge auto-heal when cwd is $HOME

ihubanov:
## What

The `sage-gui mcp` stdio bridge no longer auto-heals hooks into the user-global
config directory. `selfHealProject` returns early when the project directory
resolves to `$HOME`. Normal project heal is unchanged.

## Why

The bridge (`runMCP`, `cmd/sage-gui/mcp.go`) calls `selfHealProject` on every
startup, creating `.claude/hooks/*.sh` and merging `${CLAUDE_PROJECT_DIR}`-
relative hook commands into `.claude/settings.json`. When the bridge's working
directory is `$HOME`, `<cwd>/.claude` *is* the user-global config dir. The
`${CLAUDE_PROJECT_DIR}`-relative hooks land there and then fail in every *other*
project with a missing-file error on `Stop`/`UserPromptSubmit`.

That is the same hazard #297 guards for the explicit `install` command; the
bridge's startup self-heal is a separate code path that did not share the guard.

## How

- Guard `selfHealProject` with an early return when `projectDir` resolves to
  `$HOME`. The bridge only needs to sign requests; `mcp install` remains the
  explicit, directory-aware way to (re)install hooks.
- Extract the `$HOME` check from `errIfInstallTargetsHome` (#297) into a reusable
  `installTargetsHome(projectDir string) bool`, which fails false when home is
  unresolvable. `errIfInstallTargetsHome` now calls it — no behaviour change to
  the install guard.
- Complements #297: that PR covered the explicit `install` command; this covers
  the bridge's startup self-heal.

## CLAUDE_CONFIG_DIR is deliberately not guarded

An earlier revision also returned early when `CLAUDE_CONFIG_DIR` was set. That
was wrong: `CLAUDE_CONFIG_DIR` overrides the *user* config directory, not project
config — `projectDir/.claude/settings.json` is still discovered from the working
directory and must still heal. Guarding on it disabled valid project hook repair
(and `selfHealCodex` + CLAUDE.md repair for Codex sessions that merely inherit
the variable). Only the `$HOME` case is a genuine pollution vector, and
`projectDir/.claude` can equal `~/.claude` only when `projectDir == $HOME`.

## ABCI determinism

Consensus path untouched. `cmd/sage-gui` only; no `internal/abci`,
`internal/store`, `internal/tx`, or FinalizeBlock code involved.

## Tests

- `go test ./cmd/sage-gui/` passes; `gofmt` clean.
- `golangci-lint run` clean at the CI-pinned version.
- `TestSelfHealProject_SkipsHomeNotConfigDir` in
  `cmd/sage-gui/selfheal_guard_test.go`: heals a normal project; **heals a normal
  project even when `CLAUDE_CONFIG_DIR` is set** (the requested regression); skips
  when `projectDir == $HOME`.
- existing `TestErrIfInstallTargetsHome` still passes (shared `installTargetsHome`
  helper).

### DIFF
--- a/cmd/sage-gui/mcp.go
+++ b/cmd/sage-gui/mcp.go
@@ -513,24 +513,34 @@ func sanitizeDirName(name string) string {
 // whichever project is open — so the hooks fail in every other project. Refuse
 // rather than write a guaranteed-broken install. Fails open when $HOME cannot be
 // resolved, so it never blocks a legitimate install.
-func errIfInstallTargetsHome(projectDir string) error {
+// installTargetsHome reports whether projectDir resolves to the user's home
+// directory — where a project-scoped .claude/.codex write would land in the
+// user-global config and its ${CLAUDE_PROJECT_DIR}-relative hooks would then break
+// in every project. Fails safe (false) when home cannot be resolved, so it never
+// blocks a legitimate install/heal.
+func installTargetsHome(projectDir string) bool {
 	home, err := os.UserHomeDir()
 	if err != nil {
-		return nil
+		return false
 	}
 	projectAbs, projectErr := filepath.Abs(projectDir)
 	homeAbs, homeErr := filepath.Abs(home)
 	if projectErr != nil || homeErr != nil {
+		return false
+	}
+	return filepath.Clean(projectAbs) == filepath.Clean(homeAbs)
+}
+
+func errIfInstallTargetsHome(projectDir string) error {
+	if !installTargetsHome(projectDir) {
 		return nil
 	}
-	if filepath.Clean(projectAbs) == filepath.Clean(homeAbs) {
-		return fmt.Errorf(
-			"refusing to install into the home directory (%s): install writes a "+
-				"project-scoped config whose hooks are ${CLAUDE_PROJECT_DIR}-relative, "+
-				"which would land in the user-global config and break in every project; "+
-				"cd into a project directory and run install there", homeAbs)
-	}
-	return nil
+	home, _ := os.UserHomeDir()
+	return fmt.Errorf(
+		"refusing to install into the home directory (%s): install writes a "+
+			"project-scoped config whose hooks are ${CLAUDE_PROJECT_DIR}-relative, "+
+			"which would land in the user-global config and break in every project; "+
+			"cd into a project directory and run install there", home)
 }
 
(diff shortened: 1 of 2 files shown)


ihubanov:
CI note: the ubuntu `compile-and-contract` job failed at the "Prepare verified AppImage helper cache" step — 3 failed attempts to fetch `linuxdeploy-plugin-appimage` from GitHub releases (a transient external download), unrelated to this change. The macOS and Windows `compile-and-contract` jobs and every other check (Lint, CodeQL, Consensus Fault Gates, Analyze, etc.) are green. A re-run of that one job should clear it — I don't have re-run rights on this repo.

l33tdawg (changes requested):
The $HOME guard is sound, but the CLAUDE_CONFIG_DIR guard is not. Anthropic documents CLAUDE_CONFIG_DIR as overriding the user configuration directory (the paths that would otherwise live under ~/.claude); project configuration remains projectDir/.claude/settings.json and is still discovered from the current working directory. See https://code.claude.com/docs/en/claude-directory and https://code.claude.com/docs/en/configuration. Returning from selfHealProject whenever this variable is non-empty disables valid project hook repair, and also skips selfHealCodex plus CLAUDE.md repair when the MCP process is a Codex session that merely inherits CLAUDE_CONFIG_DIR. Please retain the projectDir == $HOME protection, remove or narrow the environment guard, and add a regression proving a normal project still heals when CLAUDE_CONFIG_DIR is set.

ihubanov:
Good catch, thanks — you're right. `CLAUDE_CONFIG_DIR` overrides the user config directory, not project config, so guarding on it wrongly disabled valid project hook repair (and `selfHealCodex` + CLAUDE.md repair for Codex sessions that merely inherit the variable).

I've dropped the `CLAUDE_CONFIG_DIR` guard and kept the `projectDir == $HOME` protection, which is the actual pollution vector — `projectDir/.claude` can only be `~/.claude` when `projectDir == $HOME`. Added the regression you asked for: `TestSelfHealProject_SkipsHomeNotConfigDir` now proves a normal project still heals when `CLAUDE_CONFIG_DIR` is set, alongside the normal-project and `$HOME`-skip cases. Reworded the commit and PR body to the `$HOME`-only rationale.

PR #328, opened 2026-09-14
v11.20.0: encrypted agent working state, durability, and Lantern bring-up

l33tdawg:
## What this ships

v11.20.0 replays the rescued feature batch onto post-v11.19.22 main as five reviewed slices, then performs the release prep. Nothing was blind-rebased: the batch was a single rescue snapshot (f55e54b1, 56 files) that had never been compiled in that state, so each slice was replayed onto main, built, tested and committed separately.

- **Storage.** SQLite opens with `synchronous=FULL` in both DSN and PRAGMA form and verifies `journal_mode`/`synchronous` at boot; vault publication is atomic (`ActivateVault`); adds an encrypted workflow journal (CAS revisions, conversation guard), a private-media JPEG store (ciphertext-only rows, actor isolation, quotas, startup disk-floor probe) and message-storage status.
- **REST.** `GET /v1/messages/storage`; `GET`/`PUT /v1/workflows[/{uuid}]`; `GET`/`PUT /v1/private-media/{uuid}`; plus a strict signed `require_encrypted_storage` send flag that fails closed with 503 instead of silently storing plaintext. All three surfaces sit inside the app-v23 pipeline agent boundary with exact no-store headers. The authenticated body limit becomes route-aware for exactly one route (canonical-UUID private-media PUT gets 2 MiB).
- **Node.** `sage-gui init-lantern-private` creates a fresh-only Lantern identity (refuses an existing or mismatched node, takes the companion-key bootstrap explicitly, installs no services and no public enrollment); `SAGE_LANTERN_PRIVATE_LISTENERS` makes a missing config an explicit error and the policy is re-checked at YAML load; a validated `private_media` config block with a startup space probe.
- **SDK and docs.** Python SDK clients/models for all three surfaces plus tests, and reference doc updates.
- **Dormant.** A public-memory Merkle index, migration builder and stage/promote path ship with tests and no production caller.

## Consensus audit

Answered from the code, not from the rescue notes: no production caller exists for `PreparePublicMemoryStage`, `PromotePublicMemoryStage`, `ValidatePublicMemoryStage`, `ComputePublicMemoryAppHash` or `SyncPublicMemoryChanges`. Staged rows live under a local namespace excluded from the AppHash; promotion is what writes into AppHash-covered `public-memory-index:v1:node:` state and is explicitly named for app-v28. This release ships that machinery **inert** and opens no fork: no consensus change, no chain migration, `internal/abci/app.go` `maxSupportedAppVersion` is untouched at 27 and is pinned by the release contract test.

## Verification (local, this branch)

- `go build ./...` clean
- `go test ./... -count=1` all packages ok (api/rest 97s, internal/store 122s, internal/abci 206s, internal/federation 178s, web 230s)
- `cd natter && go test ./...` ok
- `go vet ./...` clean
- `node --test scripts/*.test.mjs` 430/430
- `PYTHONPATH=src python3 -m pytest sdk/python/tests -q` 562 passed
- `go test -tags v119testfixture ./cmd/v119-governance-fixture/` ok
- `gofmt -l` on every ported file clean

Four doc-to-code citations drifted because the new code added lines; they were re-pointed so the citation test stays green. Port fidelity was checked by diffing the branch against the rescue snapshot: the tree carries exactly the 56 batch files and nothing else.

## Release gates that had to move together

README, ROADMAP, UPGRADING and every release-facing marker plus the `server.json` lockstep entries move to 11.20.0. Two compatibility gates move with them, because the SSCP range lives in the **shipped shell**, not the daemon: `desktop/sage-shell/src/control.rs` admits minor 20 (its boundary test now asserts 11.20.0 supported and 11.21.0 not), and `scripts/stage-native-shell-daemon.sh` accepts the 11.20 series in both the pattern and the error text, with both test pins moved in lockstep. A v11.20.0 release must ship the rebuilt shell alongside the daemon; a daemon-only upgrade would leave old shells refusing control. The federation-acceptance Compose project name uses the concatenated form `sage-v112000-federation`.

## Known, deliberate, not in this PR

PyPI publishing still sets an explicit password, so the `attestations: true` input stays ignored (tracked task 9d3edb15). Landing it before this release would make v11.20.0 the release that proves PEP 740 attestations appear on the PyPI files.


l33tdawg:
Added a sixth commit: **operator-chosen agent discovery** (`ef189c3e`).

Your framing was right and my earlier read of the code was too narrow. Memory Read is export-gated, but a trusted link advertises *every eligible ordinary local agent* to the peer for messaging: the node-messaging projection builds contacts with no domain prerequisite, and that single funnel feeds the peers listing, its paging, and its exact-name/id search. So the peer saw every agent, picked a name, and could send to the wrong one.

What landed:

- `internal/store/federated_agent_exposure.go` - a connection-scoped policy bound to the same four-part agreement generation as the sibling snapshots, with CAS revisions, a complete sorted allow list, fail-closed reads when the generation it was written against has been retired, and the documented default (`mode=all`, `configured=false`) when nothing is set. The re-pair reset retires it, so a fresh link starts at the default instead of inheriting an old allow list.
- `internal/federation/agent_exposure.go` plus one edit in the node-messaging branch of `pipe_contacts.go` - the policy filters the projection that produces the status listing, the node pages, and the lookups. Selected agents must be currently eligible ordinary agents, so nobody can allow-list a Root identity.
- `GET`/`PUT /v1/dashboard/federation/connections/{chain_id}/agent-exposure` behind the operator gate, documented in `docs/reference/rest-api.md`.
- CEREBRUM: **Agent discovery on <peer>** in the connection detail with All agents (default) / Only the ones I pick / None. Ticking the first agent narrows the connection to exactly that agent rather than adding to an empty list, later ticks add, clearing every tick means nobody, and Save commits under the revision it read. Docs and release notes updated.

Deliberate scope, stated in the UI and docs: this is discovery only. It grants no memory Read and authorises no delivery. One asymmetry worth your call - an agent whose own Access Controls policy refuses federated delivery is still advertised as "Not accepting", by design, so the peer knows it exists without being promised a route. I tried making that also hide the agent and it contradicts an existing defended contract (`TestNodeContactsRequireNoMemoryExportAndPreserveRestrictions`), so I left it; say the word if you want the deny bit to hide as well and I will change the contract deliberately.

Also fixed the two UI defects you hit, both in `.access-control-drawer`: it is a fixed-height scroller, so the identity row (with the × close button) and the save row scrolled out of reach. Both are now sticky, so the close button and **Save policy** stay visible while the body scrolls.

Verification: `go build ./...`, `go test ./...` (all packages; two web failures under parallel load were process-kill flakes in the update/restart probe tests - both pass in isolation and the whole `./web/` package is green at 171s), `node --test scripts/*.test.mjs` 430/430 after re-pointing 11 shifted doc citations, plus new store/federation/handler tests for the policy.

l33tdawg:
Two follow-ups pushed as `45fba735`:

1. **Connectome node identity.** Every node circle said "SAGE" while the machine name sat detached below the cluster. The circle now carries the node name (delimiter-aware wrapping, so STUDIO-MACMINI renders as STUDIO- / MACMINI instead of losing the hyphen), with the full value in the title/aria-label/inspector. The local node is no longer the letter S. Each node gets a colour hashed from its chain id — deterministic, nothing stored or negotiated, so both operators see a peer in the same colour — applied to the ring, the trust link, the agent dots, the halo and the inspector swatch. Colour stays redundant (names and state words are text).

2. **SDK model nit.** `MessageStorageStatus` declared a field named `schema`, shadowing `BaseModel.schema` and emitting a pydantic warning; it now uses the `schema_` + alias convention its sibling wire models already use.

Verified by rendering the actual component: `e2e/federation-connectome.spec.js` 5/5 (screenshots regenerated), `go test ./web/`, `node --test scripts/*.test.mjs` 430/430, Python SDK 562/562. The spec earned its keep — it caught a stray deleted template line that broke htm parsing, which is exactly the class of failure a string-assertion test would have missed.

l33tdawg:
Added `0c8bf600`: **the `sage_backlog` bug another agent hit.**

Diagnosis first: the tool was the only list surface here with no paging, no truncation signal, and no way to reconcile a count against a listing. It returned the entire board in one payload, so with a few dozen tasks carrying multi-KB content the client elided the middle of the JSON to fit — and because the order is stable (`created_at DESC`), the *same* slice vanished on every call. That is exactly the reported symptom: a partial board that reads as authoritative, with a count that disagrees with the listing and nothing in the response to explain it. `rg`-style paging around it did not help because `sage_list` enumerates every memory type, not tasks.

What changed:

- **Paged listing.** `limit` (default 25, max 100) and `offset`, ordered by the node order with a `memory_id` tiebreaker so pages cannot skip or repeat. Returns `total_open`, `returned`, `has_more`, `next_offset`, and the message names the next offset when a page remains.
- **Honest truncation.** `scan_capped` reports the node 500-record bound instead of hiding it, with instructions to narrow by domain or provider. `total_open` counts only what this exact agent can enumerate — a mixed-version node can hand back rows assigned to another identity, and those must not inflate the caller board size, so the server total is deliberately not used.
- **REST.** `GET /v1/memory/tasks` returns `returned` and `scan_capped` alongside the backward-compatible `total`, and both built-in stores (SQLite and Postgres) order by `created_at DESC, memory_id ASC`: `created_at` has second resolution, so a burst of tasks could otherwise page with skips or repeats.
- **Prefix resolution.** `sage_task` now accepts a unique `memory_id` prefix of 8+ hex characters and resolves it against this agent open tasks, so the documented "superseded by `958760b4`" style predecessor can actually be closed. Ambiguous prefix errors naming every match and mutates nothing; a too-short id-shaped value errors explicitly instead of becoming a confusing server 404; a non-id label passes straight through so legacy fixtures are unaffected. The result carries `resolved_from_prefix`.

Two existing tests caught real overreach in my first cut, which is why they are still green unchanged: `total_open` must not adopt a server count that includes another identity rows, and a non-UUID id must not trigger a network lookup. New tests cover whole-board paging with no skips/repeats, the scan cap, unique-prefix resolution targeting the full id, and ambiguous/short prefixes mutating nothing.

l33tdawg:
Added `fc6ed748`: **relay flapping — root cause found and fixed.**

The relay was never the problem, and neither was the hotel. Measured from the affected node, `65.108.81.134` is **301 ms RTT, 0% loss, ±6 ms jitter** — a clean but intercontinental link. The actual defect was the budget: `federationP2PCandidateTimeout = 2 * time.Second` applied to **every** candidate, direct or relayed. A Circuit Relay v2 hop costs the dialer→relay leg, the relay→peer leg and the relay own handshake before the federation TLS handshake even starts, so a relayed dial needs several round trips at ~300 ms each inside a 2 s box. That is marginal rather than broken, which is exactly the flapping signature: a message gets through in a quiet window, and every retry in between dies mid-handshake. Both nodes run the same default, so both aborted, and the peer classified its own abort as "Secure relay unavailable".

What changed:

- **Per-candidate budgets are now per path kind.** Relayed candidates get 8 s (`SAGE_FED_RELAY_CANDIDATE_TIMEOUT_MS`); direct candidates keep 2 s (`SAGE_FED_DIRECT_CANDIDATE_TIMEOUT_MS`). Chosen from the target, not applied blanket — a test proves a direct candidate at relay latency is still cut off at the direct budget.
- **Whole-operation budgets are split the same way and are tunable.** Route refresh 8 s direct / 20 s relay (`SAGE_FED_ROUTE_TIMEOUT_MS`, `SAGE_FED_RELAY_TIMEOUT_MS`; the dependent workflow budget derives from the same choice), and the CEREBRUM status probe uses 20 s instead of 6 s for a peer whose frozen candidates include a circuit target (`SAGE_FED_STATUS_TIMEOUT_MS`, `SAGE_FED_RELAY_STATUS_TIMEOUT_MS`). A relay-only peer is no longer measured with a LAN budget. All documented in `docs/reference/environment-variables.md`.
- **The verdict is no longer relay-shaped.** New transport verdicts `timeout` and `handshake_failed`, classified from the error type and text and evaluated *before* the relay verdict in both the Go and JS classifiers, with copy that names the real symptom ("No answer in time", "Connection closed during handshake"). "Secure relay unavailable" is now reserved for failures that are actually about the relay — the presence of a relay address is not an explanation of why an attempt ended. `docs/FEDERATION.md` documents the distinction.

Verified: new tests for the transport verdicts, the relay-aware budget selection, the per-candidate dial budget, and the web precedence table; `internal/federation`, `web` and `cmd/sage-gui` suites green; contract oracle 430/430 after re-pointing nine shifted citations.

Two caveats worth stating plainly: the fix helps only where the *dialing* node has it, so DKAN-TII needs the build too before its own retries stop starving the relayed candidate; and the independent, no-rebuild lever remains putting a natter relay nearer both machines — 300 ms per leg is what makes these budgets matter at all.

PR #347, opened 2026-09-18
feat(abci): raise the auto-vote ceiling to app-v28 with its evidence

l33tdawg:
The release that carries app-v28's activation evidence is the release that raises the auto-vote ceiling. `maxSupportedAppVersion` moves 27 → 28, the compiled and readiness ceilings converge, and a personal node advances its own chain across the seam with no governance ceremony — the automatic path every earlier rung took, and what v11.22.0's dormant gate was waiting for.

**Evidence carried by this change**

- `TestAppHashDeterminism_AppV28Activation` (`test/integration/apphash_determinism_appv28_test.go`, run through `deploy/scripts/run-determinism.sh`): a fresh four-validator devnet walks app-v2 → app-v28 one rung at a time with byte-identical AppHash at H-1/H/H+1 of every seam. The final rung now casts explicit votes only while the target sits above the auto-vote ceiling, so the same harness covers the dormant build and this converged one.
- Both Consensus Fault Gates on this change (`Consensus Fault Gates / App-v20 subprocess fault oracle`, and the real Comet/ABCI gate, which I am dispatching with the authorized state-sync phase).
- A promoted node surviving state-sync and restore: the real-process gate now drives its provider through app-v28 (`TARGET_APP_VERSION=28`), so a pristine receiver restores from the provider snapshot and both sides must report exact app-v28 state with converging AppHash.
- Restart and replay equivalence across H: the same gate's receiver pre-publication SIGKILL and provider SIGKILL phases, plus the replay family that pins historical blocks to their original app versions.

**What else moved**

- The fixtures that used app-v28 as the canonical unsupported target now use app-v29 (`upgrade_governance_query_test.go`, `AcquireVerifiedUpgradeSnapshotFence`, and the per-fork lockstep assertions across `internal/abci`).
- The dashboard status and `sage-gui upgrade vote` tests pin the converged reality: a ballot at the ceiling is carried by the auto-voter; only a target above it is the deliberate-vote case.
- The state-sync CI job is renamed `App-v28 real Comet/ABCI crash, partition, and state-sync gate` to match what it now drives; `docs/reference/concepts/app-v28-lifecycle.md` is marked activated with its evidence, and INDEX follows.

No transaction-type change, no chain reset, and historical blocks keep replaying under their original app versions.


l33tdawg:
Holding this: the ceiling bump cannot land until the last evidence item is green.

**What passed.** `TestAppHashDeterminism_AppV28Activation` walks app-v2 → app-v28 on a fresh four-validator devnet with byte-identical AppHash at H-1/H/H+1 of every seam, and the App-v20 subprocess fault oracle is green. In the real-process gate the chain auto-votes the whole ladder and activates app-v28 at H (no explicit votes needed once the ceiling allows it) — the automatic path this PR is about.

**What fails.** The contracted item "a promoted node surviving state-sync and restore": in `Consensus Fault Gates / App-v28 real Comet/ABCI crash, partition, and state-sync gate` the provider restarts at app-v28, boots, serves `/health` 200s, produces a couple more blocks, then stops answering its Comet RPC and the gate times out waiting for it (`RPC did not become ready`). The identical gate on main (TARGET\_APP_VERSION=27) passes, so this is v28-specific rather than a flake. A local reproduction is blocked by a macOS bind-mount ownership assertion in the same fixture, so the root cause is still open.

Until that is fixed the auto-vote ceiling must stay 27: a personal node that auto-advanced to v28 and then rebooted is exactly the shape of the failure. The activation rides the release that turns this gate green.

